PR #5, opened 2014-12-20
Shuffle the test methods in the same class(for junit4)

changgengli:
Hello,

This PR is to randomize the excution order the junit test method in the same class. I came to this idea during the work we are upgrading from java 6 to java 8. A lot of test cases were broken after upgrade because those tests are relying on the execution order which is changed after upgrade.

This is actually just a hack for several hours. I can rework on it after I get more suggestion and more understanding to ant(eg. have a parameter to turn it off). But first, would this be valuable for others?

Thanks,
Changgeng

### DIFF
--- a/WHATSNEW
+++ b/WHATSNEW
@@ -1,4 +1,4 @@
-Changes from Ant 1.9.3 TO current
+Changes from Ant 1.9.3 TO Ant 1.9.4
 ===================================
 
 Changes that could break older environments:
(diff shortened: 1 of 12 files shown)


changgengli:
Michael,

Have you got a chance to review the code?

Thanks,
Changgeng

mc1arke:
@changgengli sorry, I forgot about this and your change now doesn't merge (I think it just needs master pulled back on to your branch). If you're still interested in contributing this functionality then could you get the change back into a merge-able state and I'll give it another review for you?

PR #8, opened 2015-04-22
increase buffer size to improve transfer speed

pkures:
1024 byte buffer size is insufficient for transfering large files over faster (> 100Mbit) networks. We are transfering files tens or hundreds MB in size and transfer speed was 1MB/s. After increasing the buffer size too 100KB we are getting transfer speeds up to 60MB/s. Please consider increasing the BUFFER_SIZE in the next Ant release. Thanks, Peter

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/optional/ssh/ScpToMessage.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/ssh/ScpToMessage.java
@@ -36,7 +36,7 @@ import com.jcraft.jsch.Session;
 public class ScpToMessage extends AbstractSshMessage {
 
     private static final int HUNDRED_KILOBYTES = 102400;
-    private static final int BUFFER_SIZE = 1024;
+    private static final int BUFFER_SIZE = 100*1024;
 
     private File localFile;
     private String remotePath;


src/main/org/apache/tools/ant/taskdefs/optional/ssh/ScpToMessage.java
```diff
@@ -36,7 +36,7 @@
 public class ScpToMessage extends AbstractSshMessage {
 
     private static final int HUNDRED_KILOBYTES = 102400;
-    private static final int BUFFER_SIZE = 1024;
+    private static final int BUFFER_SIZE = 100*1024;
```
ajmas: Any reason you don't make BUFFER = HUNDRED_KILOBYTES, since that constant is defined just above?

pkures: Yes it's possible, but I didn't want to tie them together. The HUNDRED_KILOBYTES constant is used as a progress display threshold and IMHO should be named PROGRESS_DISPLAY_THRESHOLD or something like that, but that is left for another patch ;-)  I would then increase the value of PROGRESS_DISPLAY_THRESHOLD to at least 1MB, on local network 100KB file is quite small and progress display is really unnecessary.

bodewig:
I think this is something we should make configurable via a task attribute rather than hardcode new default values.

pkures:
You are right, that it should be configurable, but the default value really should be something more sensible than 1KB. There is no reason to use such small buffer, memory consumption is really not an issue here and increasing it will help most users on LANs and over internet (if they don't use dial-up connection ;-). Also, as this is my first attempt at contributing towards a major open source project I afraid to propose larger changes than changing the default value :) Also, thank you for responding. I really wasn't sure if this is the right way to propose modifications to Ant codebase.

bodewig:
github pull requests are an option that we haven't used heavily, but it is fine.  The "canonical aproach" would have been a Bugzilla issue with attachment (that's what we've been telling users for years now).

Also, if you want to get involved with Ant development (welcome!), please join the dev mailing list.

It's more about network packet size than memory.  Maybe it's a sign of me being old but 100k feels a bit too big.  Are you feeling up to the task of adding an attribute to the scp task?

pkures:
Thank you for tips how to do it more conventionally in the future. I will try to look at the code over the weekend and maybe add the attribtue to configure buffer size if I have the time to do it. 

Are you proposing to leave the default at 1KB and only allow to change it by attribute ?

When you are receiving the data over 1Gbit you need either large number of small buffers or one larger buffer. You will not handle ten thousand 1 KB packets/s from Java easiliy, I started programming on computers that had much less memory than 100 KB so I can understand your concerns  :) But again 1MB/s on such fast network is also not optimal and our current machines have gigabytes of memory.

pkures:
Also Get task uses exactly sam buffer size when downloading files, probably for the same reasons

  private static final int BIG_BUFFER_SIZE = 100 * 1024;

ajmas:
May want to look at this for a perspective: http://stackoverflow.com/questions/236861/how-do-you-determine-the-ideal-buffer-size-when-using-fileinputstream while

ajmas:
Last comment got cut short. How did you come about the 100k value? Has this been tested on networks with smaller through put? Consider the nature of data packet sizes. Not saying a bad value, though just trying to play it safe. 

Making it configurable would certainly be good. Even wondering whether there are algorithms out there for auto tuning the buffer size?

pkures:
I was thinking along the lines of ... the transfer speed is 1MB/s but theoretically on 1 Gbit network you should see up to 100MB/s transfer speed ideally, so I tried to increase 100 times. It's about how often Java gets a chance to process InputStream data and how many of those 1K packets arrive between OS context switches. 

Increasing the buffer size from 1K to 100K increased the transfer speed to 75MB/s (very large file), so the buffer size really does matter a lot. I can look how much the InputStream read method reads in one call on average - this should tell us how large the buffer needs to be for gigabit network. I can also test over our 15Mbit link, but I'am afraid that's slowest connection I have available except maybe cellular internet, hmmm, maybe I can test that also.

InputStream.read - Reads up to len bytes of data from the input stream into an array of bytes.  

In my experience it means that on slow network the only risk is that the excessive buffer space will be wasted for the duration of the transfer. As it will not wait for the buffer to fill.

But this is really getting little bit complicated for my time constraints :-) Maybe just adding the attribute to override the defaul buffer size is not a bad option after all.

bodewig:
I appreciate the offer of doing more research, but it's not necessary.

The argument about the <ftp> task made me look into the other tasks we use for transferring files over the network - <get>also uses 100k.  I'm going to merge this and #9 int the next days (unless anybody else beats me to it).

pkures:
Cool, thank you.

PR #13, opened 2016-01-02
JDK 9 drops Java 1.5 so bumped to Java 1.6

nhojpatrick:
Java 9 Early Access Build 99, builds successfully when java change to 1.6 from 1.5

Other things might also need to change, but the build is successful. JavaDoc throws issues but completes okay.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -104,8 +104,8 @@
   <property name="chmod.maxparallel" value="250"/>
   <property name="deprecation" value="false"/>
   <property name="optimize" value="true"/>
-  <property name="javac.target" value="1.5"/>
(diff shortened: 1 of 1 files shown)


bodewig:
Unfortunately this meant Ant wouldn't build on 1.5 anymore, which is still the baseline requirement for Ant 1.9.x.

It looks as if we'd be forced to drop support for 1.5 (target 1.6 means you can't even run Ant on 1.5) in order to be buildable on 1.9.

bodewig:
We've voted to keep 1.9.x as an LTS branch for Java5 compatible versions of Ant and move master to Java8 as baseline.

PR #15, opened 2016-03-03
Added JDK 9 module system support to the Java Task

tzezula:
Added main module, modulepath and moduleupgradepath into Java task.

### DIFF
--- a/src/etc/testcases/taskdefs/java.xml
+++ b/src/etc/testcases/taskdefs/java.xml
@@ -44,6 +44,9 @@
   <property name="pipeapp"
     value="org.apache.tools.ant.taskdefs.JavaTest$$PipeEntryPoint" />
 
+  <property name="module"
+    value="TestModule" />
+
   <target name="testNoJarNoClassname">
     <java/>
   </target>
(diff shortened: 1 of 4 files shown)


bodewig:
Thanks a lot Tomas, I've merged your patch.

It would be good if you could also provide a patch the java task page inside of Ant's manual.

tzezula:
Thanks Stefan, I will do a patch of Ant's java task manual.

bodewig:
Great.

Our [CI builds](https://builds.apache.org/job/Ant-Build-Matrix/) on Windows are failing because the test `testModuleAndClassnameCommandLine` expects the module to be `TestModule/org.apache.Test` while in fact it is `TestModule\org.apache.Test`. Before I start fixing the code, am I correct in assuming the separator between module and classname always is a forward slash? I'm having a hard time finding any documentation.

tzezula:
Sorry for that. I will get some Windows box for testing.
Yes, it's '/'.
The  -m option support either  <modulename>  or <modulename>/<mainclass>.

bodewig:
don't worry and thanks for your support

tzezula:
Hi Stefan
Here is a patch for the Java task page inside of Ant's manual. I've also updated the since tags to 1.9.7:
https://github.com/tzezula/ant/commit/aef407be4d33de09861a1f9677b2f15ea5ea8fdf

I've added a modulepathref attribute to the Java task to be symmetric with classpath:
https://github.com/tzezula/ant/commit/1e501f27ecc8f292ff028191e886a36d4b571d79

bodewig:
I had a few merge conflicts as I already added the since tags, but I think I've got them resolved.

Many thanks.

tzezula:
Thanks a lot Stefan!

PR #16, opened 2016-03-11
The JDK 9 module system support for the Javac Task

tzezula:
Added a support for jigsaw modulepath, upgrademodulepath and modulesourcepath to Javac task. Both single module and multi module compilation is supported.

### DIFF
--- a/src/tests/junit/org/apache/tools/ant/taskdefs/JavacTest.java
+++ b/src/tests/junit/org/apache/tools/ant/taskdefs/JavacTest.java
@@ -18,6 +18,7 @@
 
 package org.apache.tools.ant.taskdefs;
 
+import java.io.File;
 import org.apache.tools.ant.Project;
 import org.apache.tools.ant.taskdefs.compilers.CompilerAdapter;
 import org.apache.tools.ant.taskdefs.compilers.CompilerAdapterFactory;
@@ -28,10 +29,13 @@ import org.junit.Before;
 import org.junit.Test;
 
 import static org.apache.tools.ant.AntAssert.assertContains;
+import org.apache.tools.ant.BuildException;
+import org.apache.tools.ant.types.Path;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 /**
  * Testcase for <javac>.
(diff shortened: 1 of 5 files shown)


bodewig:
This looks a bit more complex than I would have naively expected, I've probably been underestimating the impact of modules.

Many thanks!

tzezula:
Yes. Unfortunately the javac `-modulesourcepath` option for the multi module compilation is more complex than existing path arguments. It's described in [JEP 261: Module System](http://openjdk.java.net/jeps/261). The important part is:

> In large systems the source code for a particular module may be spread across several different directories. In the JDK itself, e.g., the source files for a module may be found in any one of the directories src/<module>/share/classes, src/<module>/<os>/classes, or build/gensrc/<module>, where <os> is the name of the target operating system. To express this in a module source path while preserving module identities we allow each element of such a path to use braces ({ and }) to enclose commas-separated lists of alternatives and a single asterisk ( * ) to stand for the module name. The module source path for the JDK can then be written as
>{src/*/{share,<os>}/classes,build/gensrc/*}

In simple words the `javac -modulesourcepath {src/*/{share,<os>}/classes,build/gensrc/*} -d build` means that modules will be searched in the `src` or `build/gensrc` folders. The java sources of the modules located in the `src` are in folders `share/classes` or `<os>/classes` (relative to module). The java sources of the modules located in `build/gensrc` are directly in module folder. The path before the `*` in the path element is a path to module and the path after the `*` is a path in the module to the actual source root.
During the multi module compilation the javac creates module directories in the `build` folder containing the module classes.
The Ant Javac task needs to interpret the sourcepath elements to find all modules and their java files to do the up to date check.

For example in a project structure:
```
src/java.base/share/classes/
src/java.logging/share/classes/
build/java.base/
build/java.logging/
```
and javac options `-modulesourcepath src/*/{share,<os>}/classes -d build` the modules are `java.base` and `java.logging`. The `java.base` module has sources in the `src/java.base/share/classes/` and is compiled into the `build/java.base/` folder. The `java.logging` module has sources in the `src/java.logging/share/classes/` and is compiled into the `build/java.logging/` folder. The Ant Javac task needs to do up to date check of sources in `src/java.base/share/classes/` (`src/java.logging/share/classes/`) with classes in `build/java.base/` (`build/java.logging/`).

bodewig:
Many thanks for the explanation.

src/tests/junit/org/apache/tools/ant/taskdefs/JavacTest.java
```diff
@@ -244,4 +248,85 @@ public void testTargetOverridesDefault() {
...
         assertEquals("1.5", javac.getTarget());
     }
+
+    @Test
+    public void testModulesourcepathOrSrcDirRequired() {
+        try {
+            javac.checkParameters();
+            fail("Build exception should have been thrown - neither srcDir nor modulesourcepath");
+        } catch (BuildException e) {
+            //pass
```
silviu-burcea: You can use `@Test(expected = BuildException.class)` instead of useless try/catch.

bodewig: Good catch, fixed. Thanks!

tzezula: Thanks! I didn't know this.

PR #18, opened 2016-04-21
JDK9 modules support for JUnitTask

tzezula:
Changes:
1.  Added modulepath and upgrademodulepath elements.
2. When modulepath or upgrademodulepath is given VM fork is required.
3. JUnit library required by Ant is searched both on classpath and modulepath.

As seen in [JUnitTask + JDK9 question thread](http://mail-archives.apache.org/mod_mbox/ant-dev/201604.mbox/%3CAFE6C849-0622-44D1-9FF7-3A6CA4832F82%40oracle.com%3E) there are many ways how to write and execute unit test in the JDK9 involving several java options (`-addmods`, `-Xpatch`, `-XaddExports`, `-XaddReads`). 

The JunitTask can 
1. Keep the responsibility on user to correctly specify these options. The disadvantage of these solution is the complexity of VM options.
2. Automatically add the VM options. The disadvantage of these solution is that it’s impossible to cover all scenarios and there needs to be a way how to disable it. Also the options may change in time, for example if junit becomes a named module.

Currently the patch contains the 1st. solution. 
Thanks for comments!

### DIFF
--- a/manual/Tasks/junit.html
+++ b/manual/Tasks/junit.html
@@ -840,7 +840,7 @@ The <code>-XaddExports</code> java option makes the non-exported test package <c
     &lt;junit fork="true"
         jvm="${platform.java}"&gt;
         &lt;jvmarg line="-addmods ${test.module.name}"/&gt;
-        &lt;jvmarg value="-XaddExports:${test.module.name}/my.test=junit"/&gt;
+        &lt;jvmarg value="-XaddExports:${test.module.name}/my.test=junit,ALL-UNNAMED"/&gt;
         &lt;modulepath&gt;
             &lt;pathelement path="${modules}:${build.classes}:${libs.junit}"/&gt;
         &lt;/modulepath&gt;
@@ -851,7 +851,7 @@ The <code>-XaddExports</code> java option makes the non-exported test package <c
 <p>Runs my.test.TestCase as a black-box test in the forked VM given by the <code>platform.java</code> property.
 The junit library is used as an automatic module. The tests module-info requires the tested module and junit.<br/>
 The <code>-addmods</code> java option enables the test module.<br/>
-The <code>-XaddExports</code> java option makes the non-exported test package <code>my.test</code> accessible from the junit module.
-Another possibility is to export the test package in the tests module-info by <code>exports my.test</code> directive<br/>
+The <code>-XaddExports</code> java option makes the non-exported test package <code>my.test</code> accessible from the junit module and Ant's test runner.
+Another possibility is to export the test package in the tests module-info by <code>exports my.test</code> directive.<br/>
 </body>
 </html>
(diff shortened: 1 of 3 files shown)


tzezula:
I've forwarded the pull request to jigsaw-dev mailing list for comments.
http://mail.openjdk.java.net/pipermail/jigsaw-dev/2016-April/007515.html

bodewig:
Thanks so much Tomáš and sorry for taking so long to get back to you.

I may be reading the code wrong, but do we end up with the modules being on the CLASSPATH as well as the module path if JUnit ist't part of the CLASSPATH? Can this do harm?

The code

```java
return loader.getResource("junit.framework.Test") != null;
```

looks curious to me. I recall applying a fix to Ant's scriptdef subsystem because we've been told you wouldn't be able to load classes as resources anymore (we used `ClassName.class` rather than `ClassName`, though) - has this changed?

And one nit, Ant has been bitten by locale sensitive bugs before, please use `Locale.ENGLISH` when using `toLowerCase`.

bodewig:
Have you seen my comments of "22 May", @tzezula ?

tzezula:
Sorry I've overlooked the comment.

Regarding the JUnit lookup. When the JUnit is not found on the classpath it's searched on the modulepath. In fact the module path is used for 2 things here. First it's passed to the forked external process running the test(s). The presence of modulepath requires fork as there is no way how to change the module path of existing VM except of running the JUnit in a custom 
[Layer](http://hg.openjdk.java.net/jdk9/dev/jdk/file/2624d54d0103/src/java.base/share/classes/java/lang/reflect/Layer.java). The second thing the module path is used to is to lookup the JUnit needed by the JunitTask itself. In this case the JUnit is not handled as module but loaded by AntClassLoader as a jar into the unnamed module containing the Ant. I hope it should be OK.

The code: `loader.getResource("junit.framework.Test")` is clearly wrong, it should be either `loader.getResource("junit/framework/Test.class")` or `loader.loadClass("junit.framework.Test")` as done in createMirror. The `loader.getResource("junit/framework/Test.class")`  works as both the JUnit and Ant are part of the unnamed module.
Fixed  by: b9183f95e5b543acd60eaea240c6c3fc74bae748

The toLowerCase problem is fixed by: 1fc1ce108693bd52b9cf691b68dffb8f16343a42

tzezula:
I also improved the black-box test example in JUnit manual. The `addExport` to `ALL-UNNAMED` module containing the Ant is needed when Ant's test runner tries to invoke method on the test class.

bodewig:
Thanks, looks good to me. Unfortunately `git am` fails to apply the first patch to JUnitTaskTest.java:27 - any chance you could rebase your patch on the current master branch?

tzezula:
OK, I will rebase to current master.

tzezula:
Should be rebased. Let me know in case of any problems.
Thanks!

bodewig:
Unfortunately `git am` still didn't work, I've fallen back to applying patches manually (so your original commits didn't make it through).

Thanks a lot.

tzezula:
Thanks a lot Stefan!

PR #21, opened 2016-09-09
JDK9 command line options updated to JEP 293

tzezula:
The JDK9 command line options updated to [JEP-293](http://openjdk.java.net/jeps/293) - Guidelines for JDK Command-Line Tool Options.
Options introduced before the  JDK 9 will continue to exist in the old format, eg. -classpath, -cp and a new format is added: --class-path.
The options introduced in the JDK 9 were updated to the new format, eg. --module-path, -p.

This pull request updates the JDK 9 options used in the Ant to the new format.

### DIFF
--- a/manual/Tasks/junit.html
+++ b/manual/Tasks/junit.html
@@ -814,10 +814,10 @@ class.
 <pre>
     &lt;junit fork="true"
         jvm="${platform.java}"&gt;
-        &lt;jvmarg value="-Xpatch:${module.name}=${build.test.classes}"/&gt;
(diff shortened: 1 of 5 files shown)


bodewig:
Thanks.

By now I realize I should have applied your patches to the 1.9.x branch as well, will try to cherry-pick the together sometime the coming weekend.

tzezula:
Thank you Stefan.
I've overlooked one place with the old option, so I've created a new pull request #22 to fix this problem.
I've also added a patch for 1.9.x branch into #22.

PR #27, opened 2016-12-06
Bug 58661: avoid duplicate characters in stack traces

barney2k7:
This fixes https://bz.apache.org/bugzilla/show_bug.cgi?id=58661

As suspected, the problem is in the 'br-replace' template that is off by one character in some cases:
$secondhalflen is actually less than string-length($secondhalfword) in case the number of chars in word is
even. This leads to $firsthalflen being calculated too long, resulting in a duplicated character.

Note that unlike the patch attached to the bug, this pull requests solves just the duplicate character problem - nothing else.


bodewig:
Thanks a lot.

I vaguely recall looking at the original patch and thinking "this is big", but strangely never commented.

PR #30, opened 2017-01-16
Enhance scriptdef with a 'compiled' option

pyxide:
Hello,

the goal of this option is to compile the script if possible, to speed up the execution.

Proposal:
```ant
<scriptdef name="myscript" language="javascript" compiled="true"
   src="myscript.js" encoding="utf-8" />
<!-- or -->
<scriptdef name="myscript" language="javascript" compiled="true">
 // Javascript
</scriptdef>

<for ...>
  <sequential>
    <myscript/>
  </sequential>
</for>
````

This feature is implemented for `javax.script.Compilable` script engines in `org.apache.tools.ant.util.optional.JavaxScriptRunner`. The script is compiled at the first evaluation time  and is stored as a reference in the project for further evaluations. The reference name may not be adequate (`execName + ".compiledScript.0123456789"`), I guess something along the line of `MagicNames.SCRIPT_REPOSITORY = org.apache.ant.scriptrepo` would be better.

I also added an optional encoding attribute to set the encoding of the source file from the `src` attribute and modified accordingly the loading methods in `org.apache.tools.ant.util.ScriptRunnerBase`.

These attributes are both optional and the code modifications should not impact the actual behaviour of this task and the actual script evaluations when they are not present.

I developped this feature because I had a 50 Ko Javascript script run a few hundred times in an Ant script. Each loading and execution of this script took approximately 4 to 5 seconds, the whole execution took at least 10 minutes. With the help of an ant-contrib `for` loop, and scriptdef compilation, this went down to no more than 11 seconds.

Please tell me what you think,
Thank you,
Olivier


bodewig:
Many thanks, this looks good,

While looking through your changes I wonder whether we couldn't remove part of the reflection logic for master as we target Java8 and `javax.script` should be there.

pyxide:
Thank you for your positive review.

> remove part of the reflection logic for master as we target Java8 and javax.script should be there

I really don't have a strong opinion on this point. I did the first prototype with `javax.script` classes but used reflection to accommodate to the rest of the code.

While adding some test units, I discovered that the BSF manager may automatically compiles scripts.
Had the bsf library been on my Ant classpath before, I wouldn't have developped this feature.
Should we compile the scripts by default ?

On another note, I added the `encoding` option on the other script components that use
the `src` attribute (ScriptCondition, ScriptMapper, ScriptFilter, ScriptSelector).

bodewig:
WRT reflection, let's keep it the way it is for now.

I'm not a user if the `script` family so I'm not sure what the default should be. It probably costs extra time, do I benefit from compilation if the script is run exactly once?

`encoding` is a good idea, thank you.

> I guess something along the line of MagicNames.SCRIPT_REPOSITORY = org.apache.ant.scriptrepo would be better.

+1

bodewig:
@pyxide are you going to going to change this PR to use a `MagicNames` constant?

pyxide:
@bodewig, sorry for delay, the MagicNames change has now landed.

bodewig:
no worries, I wasn't sure whether you still wanted to change the code or leave it for a later fix.

Many thanks!

PR #32, opened 2017-02-06
Added option to force the csv quote char

fsteccanella:


fsteccanella:
While exporting a resultset in CSV, it can be useful wrap every column with the quote character also if not strictly necessary.

bodewig:
Thanks.

Could you please also modify the task's manual page and maybe add a test or two?

bodewig:
Many thanks!

PR #33, opened 2017-04-18
Add Orion EJB deployment tool to EjbJar taskdef.

niggemannmj:
This taskdef is used to support legacy Orion Application Server EJB deployment. We've been applying it as patch to ANT for several years now and it would simplify life to contribute the changes back to the community. 

BTW, this taskdef should also work for Oracle OC4J, but I do not have license to that product to verify it.

### DIFF
--- a/manual/Tasks/ejb.html
+++ b/manual/Tasks/ejb.html
@@ -61,6 +61,8 @@ In general these tasks are specific to the particular vendor's EJB Server.</p>
   Application Server 6.0</li>
   <li><a href="http://www.jboss.org/" target="_top">
   JBoss 2.1</a> and above EJB servers</li>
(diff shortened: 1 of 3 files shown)


bodewig:
I'd be reluctant adding new stuff for legacy containers, in particular if we can't test them ourselves. But your change looks simple enough.

Could you please also add a blurb to the documentation in `manual/Tasks/ejb.html`?

niggemannmj:
I've updated the ejb.html to add something to the documentation. It's built on top of the generic deployment tool, so the only attribute on the orion element is the destdir.

bodewig:
Many thanks, Mark!

This will become available as part of 1.9.11 and 1.10.2.

PR #34, opened 2017-06-04
eblikstad

eblikstad:
Hi

I have created a workaround for Windows regarding redirection of I/O. Several win32 console programs do not handle a closed input. Ref. https://bz.apache.org/bugzilla/show_bug.cgi?id=61153

Regards,
Espen Blikstad


janmaterne:
Thanks for the PR. The first view:
* there isnt any test (not sure it would be possible)
* baseline is Ant_1.8, which is not maintained any more.
Current Ant is 1.9 for Java5 and Ant 1.10 for Java8, see http://ant.apache.org/faq.html#java-version

PR #35, opened 2017-07-22
[1.9.x branch] Fix for BZ-43271 BZ-59648

jaikiran:
The commit here fixes the issue reported in:

https://bz.apache.org/bugzilla/show_bug.cgi?id=43271#c6
https://bz.apache.org/bugzilla/show_bug.cgi?id=59648

This commit now checks the target remote path to determine the correct remote path before issuing a `chmod` with configured `fileMode` permissions.

I have tested this _manually_ using the `scp` task but I don't have any access to systems or testcases to make sure this works for SFTP and such.

P.S: This PR is against 1.9.x branch. If necessary I can even create a separate PR for master branch.


jaikiran:
All done - applied this commit to 1.9.x branch and then merged to master branch. The pushed commit also includes updates to the WHATSNEW and contributor files.

PR #36, opened 2017-08-28
Allow faking of zip entry modification times.

GKFX:
Adds DateUtils.parseLenientDateTime. Discussed in mailing list thread "Reproducible JARs".

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/Touch.java
+++ b/src/main/org/apache/tools/ant/taskdefs/Touch.java
@@ -40,6 +40,7 @@ import org.apache.tools.ant.types.resources.FileProvider;
 import org.apache.tools.ant.types.resources.FileResource;
 import org.apache.tools.ant.types.resources.Touchable;
 import org.apache.tools.ant.types.resources.Union;
+import org.apache.tools.ant.util.DateUtils;
 import org.apache.tools.ant.util.FileNameMapper;
 import org.apache.tools.ant.util.FileUtils;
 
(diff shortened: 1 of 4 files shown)


bodewig:
Many thanks @GKFX 

I think we don't want to make `DateUtils` depend on `Touch` but rather have it the other way around. Could you please move the definition of `DEFAULT_DF_FACTORY` to `DateUtils` and have the field in `Touch` read it's value from `DateUtils`?

Tests would be nice ;-)

GKFX:
Will do, hopefully in a few days.

GKFX:
@bodewig I've done that, but `ant test` fails with the output
```
-antunit-warn-location:
     [echo] AntUnit tests must be run with /home/bateman/Geek/ant/bootstrap (or /home/bateman/Geek/ant/dist), not /usr/share/ant. Try './build.sh antunit-tests' for example.

antunit-tests:

BUILD FAILED
/home/bateman/Geek/ant/build.xml:1957: Problem: failed to create task or type antlib:org.apache.ant.antunit:antunit
Cause: The name is undefined.
Action: Check the spelling.
Action: Check that any custom tasks/types have been declared.
Action: Check that any <presetdef>/<macrodef> declarations have taken place.
No types or tasks have been defined in this namespace yet

This appears to be an antlib declaration. 
Action: Check that the implementing library exists in one of:
        -/usr/share/ant/lib
        -/home/bateman/.ant/lib
        -a directory added on the command line with the -lib argument
```
and I'm not quite sure how to fix that. Is the test I've written wrong?

janmaterne:
ant-unit-1.3 is in the repo (https://github.com/apache/ant/blob/master/lib/optional/ant-antunit-1.3.jar).
It seems that you use an Ant version for building this which does not contain this jar.

If you are building Ant itself I recommend cleaning the environment before (unset ANT_HOME) and bootstrap Ant by itself.

bodewig:
@GKFX really do as the error message says and run `./build.sh test` or `build.bat test` depending on your platform from inside the Ant source tree. The wrapper script will ensure you have all the required libraries on your classpath.

GKFX:
Thank you, that all worked great. Hopefully that's ready to merge.

bodewig:
Looks great, many thanks! I'll merge it to both branches and add `@since` markers.

We'll need to update the manual (probably of zip, jar, war and ear). Also some additional unit tests for `parseLenient` would be good (good old JUnit tests). If you want to take a stab at either, please do so in a fresh PR. Otherwise I'll carve out some time to do it myself.

GKFX:
@bodewig Great, I'll have a shot at the parseLenient tests.

PR #37, opened 2017-09-17
[1.9.x branch] BZ-60644 Fix file corruption during copy task

jaikiran:
The commit here includes a fix for the issue reported in https://bz.apache.org/bugzilla/show_bug.cgi?id=60644. As described in that issue, the copy task corrupts the source file(s) that's being copied in cases where the target file points back to the source file through symlink.

The commit here (intentionally) handles/fixes this issue in core/central location and not just in the Copy task so that all such copy operations (through the `ResourceUtils`) now have the safeguard to ensure that such copying doesn't cause the source file corruption and the copying itself is skipped if the source and destination are the same resource.

This commit also includes a testcase to verify and fix the issue.

This PR is against the 1.9.x branch. Once tested, approved and merged, I can make corresponding PR/commits to the master branch to fix it there too.

P.S: Given that this change is in a central location (the `ResourceUtils`), for obvious reasons, this PR will have to run past the complete set of tests before being merged.

### DIFF
--- a/src/main/org/apache/tools/ant/util/ResourceUtils.java
+++ b/src/main/org/apache/tools/ant/util/ResourceUtils.java
@@ -432,7 +432,7 @@ public class ResourceUtils {
                 final File sourceFile =
                     source.as(FileProvider.class).getFile();
                 try {
-                    copyUsingFileChannels(sourceFile, destFile);
+                    copyUsingFileChannels(sourceFile, destFile, project);
                     copied = true;
                 } catch (final IOException ex) {
                     String msg = "Attempt to copy " + sourceFile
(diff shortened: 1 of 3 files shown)


src/main/org/apache/tools/ant/util/ResourceUtils.java
```diff
@@ -666,6 +666,14 @@ private static void copyWithFilterSets(final Resource source, final Resource des
                                            final String outputEncoding,
                                            final Project project)
         throws IOException {
+        if (source instanceof FileResource && dest instanceof FileResource) {
```
bodewig: I think it would be better to use `FileProvider` rather than `FileResource` here (and in the next blocks). Actually I'd probably move the `instanceof` checks into `areSame` (and replace them with `resource.as(FileProvider.class)` there. This allows for other resources that know they are working on files and removes at least one line of duplicate code per `if` block :-)

bodewig:
Great change and I agree it should be fixed in the central location. Many thanks!

src/main/org/apache/tools/ant/util/ResourceUtils.java
jaikiran: That makes sense. I have changed it to follow your suggestion and updated this PR. I am going to run this against both windows on linux, through a Jenkins job, before merging this.

jaikiran:
Tests against various JDK versions against Windows and Linux went fine without any regressions. Will go ahead and merge this now.

PR #40, opened 2017-09-29
Add tests for parseLenientDateTime

GKFX:
As requested for #36.

### DIFF
--- a/src/tests/junit/org/apache/tools/ant/util/DateUtilsTest.java
+++ b/src/tests/junit/org/apache/tools/ant/util/DateUtilsTest.java
@@ -17,12 +17,13 @@
  */
 package org.apache.tools.ant.util;
 
+import java.text.ParseException;
 import java.util.Calendar;
 import java.util.TimeZone;
 
 import org.junit.Test;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
 
 /**
  * TestCase for DateUtils.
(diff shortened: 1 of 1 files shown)


bodewig:
Many thanks.

I've fixed the line-ends to contain line-feeds only as I couldn't merge your patch on a Linux box, therefore the PR now has conflicts. Sorry about that.

As written the tests depend on the timezone of the machine  running the tests:

```
Testcase: testLenientDateTime took 0,006 sec
	FAILED
expected:<1488622440000> but was:<1488618840000>
junit.framework.AssertionFailedError: expected:<1488622440000> but was:<1488618840000>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at org.apache.tools.ant.util.DateUtilsTest.testLenientDateTime(DateUtilsTest.java:119)
```

this is on GMT+2 and off by an hour.  But as this is off by an hour I think it is more about me being in daylight saving time right now while the date you picked was not (the EU switches on the last weekends of March and October).

GKFX:
That's interesting; since the values I were testing were coming out GMT, and I'm in British Summer Time at the moment, I thought the timezone was being set to UTC and so it would work everywhere. Apparently not, so I'll make a note in the documentation.

bodewig:
I haven't found the time to dig deeply into the issue, but it would be good to understand where exactly things turn into unexpected directions.

GKFX:
Changing the date to within BST does mean that, for me, it gets interpreted
as being a BST date. So, given what you see as well, I think that format
parses date/times as if they were what you had seen on a local clock. I
don't see anything in SimpleDateFormat's Javadoc that confirms that though.

PR #41, opened 2017-10-11
Fix the problem of instanceof test always return true.

BruceKuiLiu:
This instanceof test will always return true because DataType is the superclass of variable nested's class ResourceComparator. The variable nested is not initialized, it would be better to do a null test rather than an instanceof test.
http://findbugs.sourceforge.net/bugDescriptions.html#BC_VACUOUS_INSTANCEOF


bodewig:
Many thanks, merged.

I'd like to add your name to the contributors list, the `contributors,xml` file wants a first and a last name and I'm not completely sure what to put where. :-)

BruceKuiLiu:
Thanks.
First name: Kui, 
Last name: LIU.

PR #42, opened 2017-10-11
Fix the inefficient use of keySet iterator with entrySet iterator.

BruceKuiLiu:
The current source code accesses the key and value of a Hashtable entry, using a key that is retrieved from a keySet iterator.
It is more efficient to use an iterator on the entrySet of the Hashtable, to avoid the Map.get(key) lookup.
http://findbugs.sourceforge.net/bugDescriptions.html#WMI_WRONG_MAP_ITERATOR


bodewig:
I had to fix the last two commits with 61811c2bc as the code wouldn't compile otherwise. Have I done anything wrong while applying the patches?

PR #44, opened 2017-10-14
Remove the 'protected' modifier to make fields package protected.

BruceKuiLiu:
The three mutable static fields could be changed by malicious code or by accident. These fields could be made package protected to avoid this vulnerability.
http://findbugs.sourceforge.net/bugDescriptions.html#MS_PKGPROTECT


bodewig:
We've got pretty strict API backwards compatibilty rules, so we cannot change the visibility of any elements that have been part of a release already

PR #45, opened 2017-10-14
Consider returning a zero length array rather than null.

Kui-Liu:
It is often a better design to return a length zero array rather than a null reference to indicate that there are no results (i.e., an empty list of results).
This way, no explicit check for null is needed by clients of the method.
On the other hand, using null to indicate "there is no answer to this question" is probably appropriate.
http://findbugs.sourceforge.net/bugDescriptions.html#PZLA_PREFER_ZERO_LENGTH_ARRAYS


janmaterne:
While i agree and prefer returning Optional<?> or an empty List, in this case this would break backwards compatiblity. And therefore this is a no go on the public API.

asfgit:

asfgit:

asfgit:

asfgit:

PR #47, opened 2017-10-14
Add the "final" modifier to a public static field.

BruceKuiLiu:
This static field public but not final, and could be changed by malicious code or by accident from another package. The field could be made final to avoid this vulnerability.
http://findbugs.sourceforge.net/bugDescriptions.html#MS_SHOULD_BE_FINAL


bodewig:
Similar to #44 we cannot change a public API and thus will not merge this request.

In this particular case we even once changed it and had to revert the field to be non-final as we broike the Eclipse integration https://github.com/apache/ant/commit/984a03d1ceb6e4b5d194e4d639d0b0fca46d92be

PR #49, opened 2017-12-09
[master branch] - Fix BZ-58683 

jaikiran:
The commit here fixes the issue reported at https://bz.apache.org/bugzilla/show_bug.cgi?id=58683.

This commit along with fixing the issue reported in that bug, also has updated the `Symlink` task to now start relying on the Java 7 support of symlinking through the use of `java.nio.file.Files` APIs. This now allows the task to move away from spawning process(es) for dealing with symlinks and also allows this task to be functional on non-unix systems which support symbolic links.

P.S: For 1.9.x branch, I can create a separate commit, without using Java 7 APIs to fix this bug, if we do want to do that there. Let me know.

### DIFF
--- a/WHATSNEW
+++ b/WHATSNEW
@@ -17,15 +17,20 @@ Fixed bugs:
  * bootstrapping Ant on Windows failed
    Bugzilla Report 61027
 
- * Bugzilla Reports 59648 and 43271 - Fixed the issue where the 
-   SCP based tasks would try to change the permissions on the
-   parent directory of a transferred file, instead of changing it
-   on the transferred file itself.
+ * Fixed the issue where the SCP based tasks would try to change 
+   the permissions on the parent directory of a transferred file,
+   instead of changing it on the transferred file itself.
+   Bugzilla Reports 59648 and 43271
 
- * Bugzilla Report 60644 - Fixed the issue where the source file
-   being copied could end up being corrupted if the target of the
-   copy happened to be the same source file (symlinked back
-   to itself).
+ * Fixed the issue where the source file being copied could end 
+   up being corrupted if the target of the copy happened to be 
+   the same source file (symlinked back to itself).
+   Bugzilla Report 60644
+
+ * Fixed the issue where symlink creation with "overwrite=false",
+   on existing symlink whose target was a directory, would end 
+   up creating a new symlink under the target directory.
+   Bugzilla Report 58683
 
 Other changes:
 --------------
--- a/manual/Tasks/symlink.html
+++ b/manual/Tasks/symlink.html
@@ -26,10 +26,10 @@
 
 <h2><a name="symlink">Symlink</a></h2>
 <h3>Description</h3>
-<p> Manages symbolic links on Unix based platforms. Can be used to
-make an individual link, delete a link, create multiple links from properties files, 
-or create properties files describing links in the specified directories.
-Existing links are not overwritten by default.
+<p> Manages symbolic links on platforms where Java supports symbolic links.
+Can be used to make an individual link, delete a link, create multiple links 
+from properties files, or create properties files describing links in the 
+specified directories. Existing links are not overwritten by default.
 
 <p><a href="../Types/fileset.html">FileSet</a>s are used to select a
 set of links to record, or a set of property files to create links from. </p>
@@ -127,18 +127,8 @@ set of links to record, or a set of property files to create links from. </p>
   variable to an absolute path, which will remove the /some/working/directory portion
   of the above path and allow ant to find the correct commandline execution script.
  
-  <p><strong>LIMITATIONS:</strong> Because Java has no direct support for
-  handling symlinks this task divines them by comparing canonical and
-  absolute paths. On non-unix systems this may cause false positives.
-  Furthermore, any operating system on which the command
-  <code>ln -s &lt;linkname&gt; &lt;resourcename&gt;</code> is not a valid 
-  command on the command line will not be able to use action="single" or 
-  action="recreate". Action="record" and action=delete should still work. Finally, 
-  the lack of support for symlinks in Java means that all links are recorded as 
-  links to the <strong>canonical</strong> resource name. Therefore the link:
-  <code>link --> subdir/dir/../foo.bar</code> will be recorded as
-  <code>link=subdir/foo.bar</code> and restored as
-  <code>link --> subdir/foo.bar</code></p>
+  <p><strong>NOTE:</strong> Starting Ant version 1.10.2, this task relies on the symbolic
+    link support introduced in Java 7 through the java.nio.file.Files APIs</p>
 
 
 
--- a/src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java
@@ -36,9 +36,11 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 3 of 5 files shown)


WHATSNEW
```diff
@@ -27,6 +27,11 @@ Fixed bugs:
    copy happened to be the same source file (symlinked back
    to itself).
 
+ * Bugzilla Report 58683 - Fixed the issue where symlink creation
```
bodewig: I overlooked this through your earlier changes to WHATSNEW. We use to add the bugzilla issue as last line of the added entry and I think it would be good to keep a common format.

manual/Tasks/symlink.html
```diff
@@ -26,7 +26,7 @@
 
 <h2><a name="symlink">Symlink</a></h2>
 <h3>Description</h3>
-<p> Manages symbolic links on Unix based platforms. Can be used to
+<p> Manages symbolic links on platforms that support symbolic links. Can be used to
```
bodewig: rather "platform where Java supports symbolic links"? I'm sure people will yell at us as their platform supports some kind of symbolic link but Java doesn't support it.

src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java
```diff
@@ -207,26 +198,30 @@ public void recreate() throws BuildException {
...
+                } catch (IOException e) {
+                    if (failonerror) {
+                        throw new BuildException("Failed to recreate symlink " + lnk + " to target " + res, e);
+                    }
+                    // log and continue
+                    log("Failed to recreate symlink " + lnk + " to target " + res, Project.MSG_INFO);
+                    continue;
                 }
+                // create the link
+                this.doLink(res, lnk);
```
bodewig: I think this should be inside of the `try` block.

bodewig: Just now see `doLink` doesn't throw any `IOException`. In this case the error message is wrong as the error occured while trying to check whether it is a (self-)link.

src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java
```diff
@@ -448,49 +432,67 @@ private void handleError(String msg) {
...
+        // file already exists, see if we are allowed to overwrite
+        if (!overwrite) {
+            log("Skipping symlink creation, since file at " + lnk + " already exists and overwrite is set to false", Project.MSG_INFO);
+            return;
+        }
+        // we have been asked to overwrite, so we now do the necessary steps
+
+        // initiate a deletion *only* if the path is a symlink, else we fail with error
+        if (!Files.isSymbolicLink(link)) {
+            throw new BuildException("Cannot overwrite, as symlink, at " + lnk + " since the path already exists and isn't a symlink");
```
bodewig: should this take `failOnError` into account?

src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java
```diff
@@ -500,18 +502,12 @@ private void doLink(String res, String lnk) throws BuildException {
...
-                        String name = f.getName();
-                        if (SYMLINK_UTILS.isSymbolicLink(pf, name)) {
-                            result.add(new File(pf.getCanonicalFile(), name));
-                        }
-                    } catch (IOException e) {
-                        handleError("IOException: " + path + " omitted");
+                    .flatMap(Stream::of).forEach(path -> {
+                    final File f = new File(dir, path);
+                    if (Files.isSymbolicLink(f.toPath())) {
+                        result.add(f);
```
bodewig: The previous code canonicalized the link's parent directory to deal with symlinks further up the tree, I think this is important for consumers of the returned set like `record` that so merges entries from several "directories" that actually are links to the same directory.

To be honest I'm not sure about the real impact but I guess we better keep the existing behavior.

src/main/org/apache/tools/ant/util/SymbolicLinkUtils.java
```diff
@@ -30,6 +30,8 @@
  * a symbolic link based on the absent support for them in Java.
  *
  * @since Ant 1.8.0
+ * @deprecated Starting Ant 1.10.2, this utility is deprecated in favour of the symlink
+ *              support introduced in Java {@link java.nio.file.Files} APIs
```
bodewig: the class is used all over the code base (well, at least in a few different places :-). Do you intend to change those other occurrences as well?

bodewig:
Thanks @jaikiran, I've added a buch of inline comments.

Overall I'm in favor of this change and if you want to spend the time on fixing the bugzilla issue in a Java5 friendly way for 1.9.x that would certainly be good - bit not something I'd see as an requirement.

The `SymlinkTest` still ensures tests are skipped on platforms that are not Unix (see the `assume` in `setup`). I'm not sure if we can figure out whether Java supports symlinks on the platform easily but it would certainly be good to test things on non-Unix platforms if we claim the task should work.

WHATSNEW
jaikiran: Fixed and updated the PR.

manual/Tasks/symlink.html
jaikiran: Fixed and updated the PR

src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java
jaikiran: You are right - the error message was incorrect. I have fixed that and updated the PR.

src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java
jaikiran: Yes, that was an oversight. Fixed.

src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java
jaikiran: When I initially changed this part, while submitting the PR, I had thought about it a bit whether or not to stick with the previous behaviour. Given that it was noted as a "limitation" (in the javadoc), I had decided to change the behaviour. But thinking about it again, I do agree with you that it isn't worth changing the previous behaviour (without knowing the complete impact). So I have switched back to the previous semantic, in this part of the code and updated the PR

src/main/org/apache/tools/ant/util/SymbolicLinkUtils.java
jaikiran: >> Do you intend to change those other occurrences as well?

I do intend to do necessary changes related to this, in subsequent commit(s), when I get a chance. I didn't want to create one large PR with too many unrelated changes. I added this deprecation note just so that any new code doesn't end up using this class, now that we are on Java 7. But I think, it's better to add this deprecation note when I do get to changing references to this class in some other commits. So I have undone this change and updated the PR.

jaikiran:
>> Overall I'm in favor of this change and if you want to spend the time on fixing the bugzilla issue in a Java5 friendly way for 1.9.x that would certainly be good - bit not something I'd see as an requirement.

Given that this bug isn't a blocker or (relatively) that critical, I most likely won't backport this to 1.9.x,  unless of course there's a genuine demand to do that (and that the user cannot upgrade to 1.10.x).

src/main/org/apache/tools/ant/util/SymbolicLinkUtils.java
bodewig: sounds good to me, thanks. I prefer small changes myself as well.

bodewig:
Looks good to me, thanks!

WRT a Java5 compatible backport, I'd leave this really up to you. I don't see a pressing need for backporting either.

jaikiran:
Committed to master branch https://github.com/apache/ant/commit/cefdbd398d8e310b218f9e2ca6f0b7fb13eddbb9

PR #52, opened 2017-12-16
[master] Fix BZ-58451 BZ-58833

jaikiran:
The commit here fixes the issues reported in:

https://bz.apache.org/bugzilla/show_bug.cgi?id=58833
https://bz.apache.org/bugzilla/show_bug.cgi?id=58451

The `PumpStreamHandler` and the `StreamPumper` work together for redirecting the streams (reading from inputstream and writing to a outputstream). When the `StreamPumper` is asked to stop, it stops reading from the input stream and goes on to finish any (non-blocking) reads and finally a (potentially blocking) flush of whatever it has read so far. The `PumpStreamHandler`, which initiates that `stop` is a bit too impatient (mainly because of its inability to know what's going on in `StreamPumper` after the `stop` was invoked on it) and starts interrupting the `StreamPumper`'s thread very soon (as early as 200 milli seconds after stop). This triggers a bunch of issues in `StreamPumper`, which either is in the middle of finishing the non-blocking reads or in the middle of actually flushing out whatever it's held on to and ultimately leads to a non-clean stop and thus corrupting/truncating the output stream as noted in those issues.

The commit here introduces a way, through the `PostStopHandle`, for `PumpStreamHandler` to be aware that upon `stop` the `StreamPumper` has respected the call to `stop` and is doing certain post-stop finishing acts before actually being considered `finished`. The `PumpStreamHandler` then allows a bit of time for the post-stop tasks to finish before actually trying to kill off the `StreamPumper` using thread interrupts, if the `StreamPumper` didn't finish after that grace period.

The commit here intentionally does _not_ introduce a configurable "grace period" for the post-stop activities and instead using a hard coded (maximum) 2 seconds for the following reasons:

* The 2 seconds is the _maxium_ amount of time to let the post-stop to complete, so it's not always going to wait that long

* The cleaning up gracefully is an internal implementation detail and doesn't have to end up being configured by the user.

* Finally, the `PumpStreamHandler` anyway falls back to the thread interrupts if the post-stop (flush() etc...) doesn't complete in those 2 seconds. So from a end user point of view, there isn't really any behaviour change, except that we now give the `StreamPumper` a chance to gracefully stop.

This PR is against master branch. If this is approved I'll backport it to 1.9.x branch too.

### DIFF
--- a/src/etc/testcases/taskdefs/exec/exec-with-redirector.xml
+++ b/src/etc/testcases/taskdefs/exec/exec-with-redirector.xml
@@ -1,0 +1,122 @@
+<?xml version="1.0"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project name="exec-redirector-test" basedir=".">
+
+    <target name="setUp">
+        <!-- This "output" property is set on the project in the Java test case (ExecStreamRedirectorTest) -->
+        <mkdir dir="${output}"/>
+        <condition property="dir.listing.command" value="ls" else="cmd.exe">
+            <os family="unix"/>
+        </condition>
+        <condition property="dir.to.ls" value="/usr/bin" else="${user.dir}">
+            <os family="unix"/>
+        </condition>
+        <condition property="dir.listing.command.arg" value="-l" else="dir">
+            <os family="unix"/>
+        </condition>
+        <property name="dir.to.ls" value="/usr/bin"/>
+    </target>
+
+    <target name="list-dir">
+        <!-- Just do listing of the same directory and redirect the output to different files -->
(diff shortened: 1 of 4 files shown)


asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/14/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/8/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/15/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/9/

jaikiran:
Have updated the PR to fix the test case so that it runs the correct command on Windows, during the tests.

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/16/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/10/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/15/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/21/

jaikiran:
@bodewig, do you think this change is OK? Any thoughts/concerns? I understand that this is one of the central classes, so don't want to push this if there's any concerns with this change.

bodewig:
I must admit that I'm scared of every change in this area, we've found and fixed too many subtle bugs in code that looked right.

Your change looks very reasonable and if all tests pass on Windows as well, then I'm fine with it (but still scared :-).

jaikiran:
Thanks Stefan (@bodewig) for the review. The tests (existing and a new one in this PR) are passing on both *nix and Windows. Given that this change does address the issues noted in the referenced bugzilla issues (tested both manually and automated), I will go ahead and merge this change and keep a watch on any issues in and around this area if/when they get reported.

jaikiran:
This is now commit to both master and 1.9.x branches https://github.com/apache/ant/commit/6c860e0036a8742f4cc35e3e6a8e595e42804c89

PR #53, opened 2017-12-19
Optional libraries for Ant 1.10

twogee:
A bit unsure about JRuby: should we stay with 1.7 for a while or go for 9? In any case, would jruby-core be sufficient, or should jruby-stdlib be added, too?

### DIFF
--- a/WHATSNEW
+++ b/WHATSNEW
@@ -17,10 +17,11 @@ Changes that could break older environments:
    should instead be changed to use resource collections like:
 
    <delete>
-   	 <fileset dir="/foo/bar/" includes="*.something"/> 
+   	 <fileset dir="/foo/bar/" includes="*.something"/>
    </delete>
 
-
+ * Commons Net 3.6 is binary-code, but not source compatible;
+   see change list of Commons Net 3.0 for details
 
 Fixed bugs:
 -----------
@@ -48,7 +49,7 @@ Fixed bugs:
    Bugzilla Report 58683
 
  * Improvement to the Zip task for reduced memory usage in certain
-   cases. Thanks to Glen Lewis for reporting the issue and 
+   cases. Thanks to Glen Lewis for reporting the issue and
    suggesting the fix.
    Bugzilla Report 19516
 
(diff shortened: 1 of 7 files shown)


asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/14/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/20/

bodewig:
We only add what we need on order to get things compiled - or in the case of jruby in order to run the tests. If what we do works without stdlib I don't see any reason to add it. People who want to use JRuby with Ant will add the dependency they need themselves.

The Commons Net change looks as if it should be mentioned as a potentially breaking change.

twogee:
I assumed that the intention of fetch.xml is to provide something usable in runtime rather than at compile time. Nevertheless, I'd rather leave JRuby as it is, since the newer versions have a lot of dependencies.

WRT Commons Net, what kind of changes may be troublesome? The change log states that 3 is binary compatible with 2 (some extraneous exceptions were removed). While we're at that, perhaps 1.9.x branch should use 2.2 (the official Java 5 release)?

bodewig:
Ahh, there is the misunderstanding. `fetch.xml` is only there for Ant development, not for Ant users at all. It is there to be used by people who want to build and test Ant. Users are expected to pick their dependencies as they need them.

I'm not sure whether there are breaking changes between Commons Net 1.x and 3.x but if Ant is compiled against 3.x chances are it won't work against 1.x anymore, which would break the builds of users who depend on 1.x (and never cared about `fetch.xml` at all as they managed their dependencies themselves).

twogee:
IMHO, the whole point with `-Ddest` in `fetch.xml` is being able to complement `$ANT_HOME/lib` in Ant distribution. Otherwise, we may just use Ivy and put the dependencies (except NetREXX) into `ivy.xml`

Any problems with using 2.2 for 1.9.x?

bodewig:
I see you've already merged this PR, I'll look st it later. I hope you didn't wait for me to respond during Christmas.

I'm not sure why `dest` is part of `fetch.xml`, we've always seen at as something for the Ant developers and it really only sets the baseline Ant is compiled against. No more.

I'm not sure which benefit you see with upgrading all those libraries as from my point of view you are only updating what the binary distributions get compiled against (and only if the release manager deletes the already existing local cache and re-runs `fetch.xml` which is rather unlikely to happen). Either the API of the upgraded version remains compatible with the older one so the update doesn't have any effect at all or it is not, in which case we introduce a backwards incompatibility with no real win I can see. But that's just my opinion.

There is a page inside of the manual - http://ant.apache.org/manual/install.html#librarydependencies - that may need to get updated as well.

twogee:
I see `dest` as a useful functionality of fetch.xml (why else should it be a part of the distribution?); and that is basically the crux of the matter.

I will update the documentation, thanks for the pointer.

PR #58, opened 2018-02-05
Use StringBuilder instead of StringBuffer as it offers high performan…

reudismam:
…ce in single thread places as it is generally the case.

### DIFF
--- a/src/main/org/apache/tools/ant/IntrospectionHelper.java
+++ b/src/main/org/apache/tools/ant/IntrospectionHelper.java
@@ -1662,7 +1662,7 @@ public final class IntrospectionHelper {
             return text;
         }
         final int ends = (MAX_REPORT_NESTED_TEXT - ELLIPSIS.length()) / 2;
-        return new StringBuffer(text).replace(ends, text.length() - ends, ELLIPSIS).toString();
+        return new StringBuilder(text).replace(ends, text.length() - ends, ELLIPSIS).toString();
     }
 
 
--- a/src/main/org/apache/tools/ant/Main.java
+++ b/src/main/org/apache/tools/ant/Main.java
@@ -1089,7 +1089,7 @@ public class Main implements AntMain {
                 in.close();
                 shortAntVersion = props.getProperty("VERSION");
 
-                final StringBuffer msg = new StringBuffer();
+                final StringBuilder msg = new StringBuilder();
                 msg.append("Apache Ant(TM) version ");
                 msg.append(shortAntVersion);
                 msg.append(" compiled on ");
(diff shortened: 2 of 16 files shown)


asfgit:

asfgit:

src/main/org/apache/tools/ant/listener/MailLogger.java
```diff
@@ -102,7 +102,7 @@
     private static final String DEFAULT_MIME_TYPE = "text/plain";
 
     /** Buffer in which the message is constructed prior to sending */
-    private StringBuffer buffer = new StringBuffer();
+    private StringBuilder buffer = new StringBuilder();
```
jaikiran: I'm not too sure this change here, in this class is correct. The `StringBuffer` is a thread-safe class whereas the `StringBuilder` isn't. Having said that I haven't checked yet whether MailLogger class is expected to be thread safe nor have I checked its usage thoroughly.

src/main/org/apache/tools/ant/taskdefs/Parallel.java
```diff
@@ -377,7 +377,7 @@ public synchronized void run() {
         }
 
         // now did any of the threads throw an exception
-        exceptionMessage = new StringBuffer();
+        exceptionMessage = new StringBuilder();
```
jaikiran: Same comment as above.

jaikiran:
Overall, IMO, this can't be a general find/replace change and instead needs to be checked for individuals places where the `StringBuffer` is used and whether the change to `StringBuilder` will impact any thread safety guarantees if any.

src/main/org/apache/tools/ant/listener/MailLogger.java
bodewig: Given the `parallel` task and things like our execute framework that spawns new threads, loggers need to be thread safe, But to be honest I don't think we've ever verified `MailLogger` actually is.

src/main/org/apache/tools/ant/taskdefs/Parallel.java
bodewig: Looking through the class I don't think `exceptionMessage` needs to be an instance field at all, it could be a local variable in `spinThreads` and get passed as an argument to `processExceptions` without doing any harm. To me it seems it is only ever used by a single thread.

Most probably a further refactoring could get rid of the `first*` instance fields as well and have `processExceptions` return all their values nicely encapsulated in a single object - including the accumulated messages.

reudismam:
Undo edits to src/main/org/apache/tools/ant/listener/MailLogger.java and src/main/org/apache/tools/ant/taskdefs/Parallel.java

bodewig:
The changes to `FixCRLF`, `TaskOutputStream`,  `VerifyJar`, `Message`, `TraxLiaison` and `RegexpPatternMapper` all change instance variables of classes that might be used in a mutlithreaded context. In the case of `RegexpPatternMapper`  they are even protected and thus the change would break backwards compatibility.

Please restrict your change to `StringBuffer`s created as local variables in methods that are not returned or passed to multiple threads.

PR #60, opened 2018-02-15
JUnit 5 support - A new junitlauncher task

jaikiran:
This is the initial working version of a new `junitlauncher` task that support using JUnit 5 framework for testing, within Ant build files. The commit in this PR is the initial set of goals that I had in mind for the first release of this task.

The manual for this task can be (temporarily) found at https://home.apache.org/~jaikiran/temp_workspace/manual/Tasks/junitlauncher.html

The dev list discussion about this PR can be found at https://www.mail-archive.com/dev@ant.apache.org/msg46656.html


asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/40/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/34/

jaikiran:
Please ignore the Jenkins failures on this PR. I'll sort out that part in a separate discussion.

jaikiran:
retest this please

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/35/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/41/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/38/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/44/

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    @Override
+    public void sysErrAvailable(final byte[] data) {
+        if (this.sysErrStream == null) {
+            try {
+                this.sysErrFilePath = Files.createTempFile(null, "syserr");
+                this.sysErrFilePath.toFile().deleteOnExit();
+                this.sysErrStream = Files.newOutputStream(this.sysOutFilePath);
```
bodewig: copy-paste error, you want that to be `sysErrFilePath` :-)

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
```diff
@@ -0,0 +1,139 @@
...
+        return Optional.empty();
+    }
+
+    private void writeFrom(final Path path, final Writer writer) throws IOException {
+        final byte[] content = new byte[1024];
+        int numBytes;
+        try (final InputStream is = Files.newInputStream(path)) {
+            while ((numBytes = is.read(content)) != -1) {
+                writer.write(new String(content, 0, numBytes));
+            }
```
bodewig: I don't think this is going to work reliably. The `read` may have split a multi-byte sequence at the end of `content` and then creating a string from it is going to break. Is there any reason you want to use a stream when reading the temporary file rather than a reader?

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Override
+    public void close() throws IOException {
+        if (this.sysOutStream != null) {
+            try {
+                this.sysOutStream.close();
+            } catch (Exception e) {
+                // ignore
+            }
+        }
```
bodewig: `FileUtils.close`?

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
jaikiran: Indeed :) Fixed.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
jaikiran: You are right, good catch! I hadn't considered that scenario. I have now switched to using a reader instead.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
jaikiran: Yes, make sense. Fixed.

jaikiran:
@bodewig, Thank you very much for the review. Based on the review comments, I have updated the PR with additional separate commits (for easier reference). I'll be squashing all these commits together during merge.

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/45/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/39/

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
```diff
@@ -0,0 +1,508 @@
...
+            Thread.currentThread().setContextClassLoader(previousClassLoader);
+        }
+    }
+
+    /**
+     * @return Creates and returns the a {@link Path} which will be used as the classpath of this
+     * task. This classpath will then be used for execution of the tests
+     */
+    public Path createClassPath() {
+        this.classPath = new Path(getProject());
```
bodewig: if users specify multiple nested classpath elements, only the last one will be used. You may want to return an already created `this.classpath` or move on to an `add` method and collect the paths inside of a wrapper `Path`.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
```diff
@@ -0,0 +1,508 @@
...
+        return this.classPath;
+    }
+
+    /**
+     * @return Creates and returns a {@link SingleTestClass}. This test will be considered part of the
+     * tests that will be passed on to the underlying JUnit platform for possible execution of the test
+     */
+    public SingleTestClass createTest() {
+        final SingleTestClass test = new SingleTestClass();
+        this.preConfigure(test);
```
bodewig: `create*` is called before the attribute setters, If you call `preconfigure` here the test won't see the actual values set at the task level.

`addConfigured` would get called after the attribute setters, alternatively you need to defer pre-configuration until `execute` is called. With both approaches you need to ensure you don't overwrite the test's explicit configuration with values from the task, of course.

Same for `createTestClasses`.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
```diff
@@ -0,0 +1,508 @@
...
+        sysErrStreamer.setDaemon(true);
+        sysErrStreamer.setName("junitlauncher-syserr-stream-reader");
+        sysErrStreamer.setUncaughtExceptionHandler((t, e) -> this.log("Failed in syserr streaming", e, Project.MSG_INFO));
+        sysErrStreamer.start();
+        return Optional.of(new SwitchedStreamHandle(pipedOutputStream, streamer));
+    }
+
+    private static void safeClose(final Closeable... closeables) {
+        for (final Closeable closeable : closeables) {
+            try {
```
bodewig: `FileUtils.close`?

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
```diff
@@ -0,0 +1,508 @@
...
+            final Thread deliveryThread = new Thread(streamContentDeliver);
+            deliveryThread.setName("junitlauncher-" + (this.streamType == StreamType.SYS_OUT ? "sysout" : "syserr") + "-stream-deliverer");
+            deliveryThread.setDaemon(true);
+            deliveryThread.start();
+            this.contentDeliverer = streamContentDeliver;
+            int numRead = -1;
+            final byte[] data = new byte[1024];
+            try {
+                while ((numRead = this.sourceStream.read(data)) != -1) {
+                    final byte[] copy = new byte[numRead];
```
bodewig: `Arrays.copyOf`?

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
```diff
@@ -0,0 +1,508 @@
...
+            this.task = task;
+            this.sourceStream = source;
+            this.streamType = streamType;
+            this.resultFormatters = resultFormatters == null ? Collections.emptyList() : resultFormatters;
+        }
+
+        @Override
+        public void run() {
+            if (this.resultFormatters.isEmpty()) {
+                // no one to feed the stream content to
```
bodewig: I believe this may cause the `PipedOutputStream` to consider the `PipedInputStream` "broken" soon and throw on any `write` operation.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
jaikiran: You are right. I hadn't taken that into account. I have updated the PR to change this part of the code.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
jaikiran: Thank you, for catching this. I went back and read the Ant manual about writing custom tasks (which I had skipped the first time around) and now understand this logic of execution better. I have updated the PR to fix this and use the `addConfiguredXXX` construct and also made sure my `preConfigure` implementation doesn't overwrite values that might have been specified on the nested elements.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
jaikiran: Fixed.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
jaikiran: Done

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/46/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/40/

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
jaikiran: This check is actually "dead code", in the sense that this will never be true. I wanted to avoid running these threads when there's no result formatter interested in the sysout/syserr content. But that check obviously needs to happen before the thread is even created and in fact, there's already such a check in the `trySwitchSysOut` and `trySwitchSysErr` methods (the place where this thread gets created).

So I've now updated the PR to remove this check.

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/47/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/48/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/42/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/43/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/49/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/44/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/45/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/50/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/51/

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
```diff
@@ -0,0 +1,295 @@
...
+            // we haven't yet created a file store and the data can fit in memory,
+            // so we write it in our buffer
+            try {
+                this.inMemoryStore.put(data);
+                return;
+            } catch (BufferOverflowException boe) {
+                // the buffer capacity can't hold this incoming data, so this
+                // incoming data hasn't been transferred to the buffer. let's
+                // now fall back to a file store
+                this.usingFileStore = true;
```
bodewig: maybe defer setting the flag until `storeToFile` below succeeds?

Also, is there any chance this method could be invoked concurrently? I haven't fully checked every execution path so maybe you are synchronizing somewhere. Tests could spawn new threads you are not aware of and these threads could be writing to syserr/out concurrently.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
```diff
@@ -0,0 +1,295 @@
...
+            if (this.fileOutputStream == null) {
+                // no backing file was created so we can't do anything
+                return;
+            }
+            this.fileOutputStream.write(data, offset, length);
+        }
+
+        private FileOutputStream createFileStore() throws IOException {
+            this.filePath = Files.createTempFile(null, this.tmpFileSuffix);
+            this.filePath.toFile().deleteOnExit();
```
bodewig: `FileUtils.createTempFile`? Probably not worth it.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
```diff
@@ -0,0 +1,508 @@
...
+        final SysOutErrStreamReader streamer = new SysOutErrStreamReader(this, pipedInputStream,
+                StreamType.SYS_ERR, testRequest.getSysErrInterests());
+        final Thread sysErrStreamer = new Thread(streamer);
+        sysErrStreamer.setDaemon(true);
+        sysErrStreamer.setName("junitlauncher-syserr-stream-reader");
+        sysErrStreamer.setUncaughtExceptionHandler((t, e) -> this.log("Failed in syserr streaming", e, Project.MSG_INFO));
+        sysErrStreamer.start();
+        return Optional.of(new SwitchedStreamHandle(pipedOutputStream, streamer));
+    }
+
```
bodewig: you could probably DRY up the two methods a little by extracting the common logic.

bodewig:
Nice work!

I don't see any reason not to merge this PR.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
jaikiran: >> maybe defer setting the flag until storeToFile below succeeds?
I did actually think of it while coding this part. The reason I decided to go this way, was to make sure that as soon as the memory size limit is reached, I prefered it to be better to no more use the in memory store even if the file store cannot be created/used.

That would imply that the subsequent syserr/sysout message will be lost but at least that won't lead to the in memory store going past the limit that's set.

The alternate approach would be to let the in-memory store be used even past its limit (and not lose any sysout/syserr content) if the file store cannot be created for whatever reason. If that feels more better approach to take, I can switch this logic in here.

jaikiran: >> Also, is there any chance this method could be invoked concurrently? I haven't fully checked every execution path so maybe you are synchronizing somewhere. Tests could spawn new threads you are not aware of and these threads could be writing to syserr/out concurrently.

One thing I didn't note in the implementation comments is that the `sysOutAvailable` and `sysErrAvailable` on these result formatter implementations will always be invoked by a single thread - the thread that we create and control `SysOutErrContentDeliverer`. Multiple threads write out to our redirected sysout/syserr stream which then gets read by the `SysOutErrStreamReader` thread which then hands off that read data to the `SysOutErrContentDeliverer` thread to have it delivered to these result formatters.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
jaikiran: I hadn't noticed that API. I think it makes sense to use it for consistency.

jaikiran:
Thank you Stefan for the thorough review. I'll clean up the commits to squash them into one and do these minor final changes before merging.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTask.java
jaikiran: Yes makes sense, I'll clean up this part a bit.

jaikiran:
This is now merged upstream. Closing this PR. Thanks Stefan for the extensive review.

PR #67, opened 2018-08-05
Support for fork mode in junitlauncher

jaikiran:
The commit here introduces support for launching the JUnit tests, through `junitlauncher` task, in forked mode. 

I decided to keep the fork aspects as a separate element instead of introducing multiple attribtues that are only applicable in forked mode. As such, each `test` or `testclasses` element of the `junitlauncher` task can now have a nested `fork` element indicating that those tests need to be launced in a forked JVM. The characteristics of the forked JVM are determined by the attribtues and nested elements of the `fork` element. 

I have added support for most of the fork attribtues that are applicable for the legacy junit task (I also checked the java task to make sure the JVM launching characteristics are captured in the fork element's attributes). I am working on the manual updates to explain this support so this PR doesn't include that part, but here's what it will end up looking like:
```
<junitlauncher>
        <classpath refid="test.classpath"/>
        <test name="org.example.junitlauncher.vintage.JUnit4SampleTest" outputdir="${output.dir}">
            
            <fork dir="${basedir}" timeout="..." 
                    includeJunitPlatformLibraries="..." 
                    includeAntRuntimeLibraries="...">

            	<sysproperty key="foo" value="bar"/>
            	...
            	<env key="hello" value="world"/>
            	...
            	<modulepath>
		     <pathelement location="lib"/>
		     <pathelement location="dist/foo.jar"/>
		</modulepath>
		...
            	<upgrademodulepath>
		    <pathelement location="upgrades"/>
		</upgrademodulepath>
            </fork>
            ...
            <listener type="legacy-xml" sendSysErr="true" sendSysOut="true"/>
            ...
        </test>
    </junitlauncher>
```
From an implementation detail point of view, the core logic of launching the tests through the JUnit platform remains intact (of course, the code itself has been moved into an internal class to be shared in forked and non-forked mode). An internal contract `LaunchDefinition` has been introduced so that the launching aspects can be captured in this interface. Non-forked and forked mode execution will internally construct an instance of the `LaunchDefinition`. 

A `StandaloneLauncher` (an internal detail of this task) has been introduced to be the entry point with a `main` method for forked mode execution. The responsibility of this class is to parse the arguments and construct the `LaunchDefinition` and then just pass it over to the `LauncherSupport` (interal impl detail). We pass around the launch definition to the forked mode launcher in the form of an xml which captures the necessary details like what tests to launch and what listeners to use. Note that this xml is an internal detail and can change over releases. I decided to capture these details in a file and pass it to the `main` method instead of passing mutliple different arguments for two reasons:

  - Reduce the command line length when executing these forked tests
  - Allow hierarchical representation of the launch definition details, like which listener is for which test.

I have tested the fork support manually, but this needs more automated tests. I'm in the process of writing those tests and also updating the manual of this task. I wanted to get any review comments on these changes in the meantime.

### DIFF
--- a/src/etc/testcases/taskdefs/optional/junitlauncher.xml
+++ b/src/etc/testcases/taskdefs/optional/junitlauncher.xml
@@ -30,6 +30,8 @@
 
     <path id="junit.engine.vintage.classpath">
         <fileset dir="../../../../../lib/optional" includes="junit-vintage-engine*.jar"/>
+        <fileset dir="../../../../../lib/optional" includes="junit-*.jar"/>
+        <fileset dir="../../../../../lib/optional" includes="hamcrest*.jar"/>
     </path>
 
     <path id="junit.engine.jupiter.classpath">
@@ -107,6 +109,16 @@
                 <listener type="legacy-brief" sendSysOut="true"/>
                 <listener type="legacy-xml" sendSysErr="true" sendSysOut="true"/>
             </testclasses>
+        </junitlauncher>
+    </target>
+
+    <target name="test-basic-fork" depends="init">
+        <junitlauncher>
+            <classpath refid="test.classpath"/>
+            <test name="org.example.junitlauncher.vintage.JUnit4SampleTest" outputdir="${output.dir}">
+                <fork dir="${basedir}"/>
+                <listener type="legacy-xml" sendSysErr="true" sendSysOut="true"/>
+            </test>
         </junitlauncher>
     </target>
 </project>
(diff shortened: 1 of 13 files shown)


asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/72/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/78/

bodewig:
Looks good at first glance, I'll have a more thorough look sometime before we cut the next release.

The new files need to have license headers added.

jaikiran:
>> Looks good at first glance, I'll have a more thorough look sometime before we cut the next release.

Thank you.

>> The new files need to have license headers added.

Will add them and setup my IDE correctly.

PR #80, opened 2018-12-07
Added tasks for JDK's jmod and jlink tools.

craigpell:
Support for the jmod and jlink tools present in the JDK since Java 9.  Now that Java 11 has no standalone JRE, officially, these tools are the only way to distribute client-side Java applications.


asfgit:

asfgit:

jaikiran:
this is ok to test

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/101/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/95/

craigpell:
All done, except that I’m finding there aren’t many man pages on my Linux system which aren’t GPL.  I’ve found a few which are under a BSD license;  is that acceptable?

src/main/org/apache/tools/ant/taskdefs/modules/Jmod.java
```diff
@@ -0,0 +1,1282 @@
...
+        }
+
+        ToolProvider jmod = ToolProvider.findFirst("jmod").orElseThrow(
+            () -> new BuildException("jmod tool not found in JDK.",
+                getLocation()));
+
+        log("Executing: jmod " + String.join(" ", args), Project.MSG_VERBOSE);
+
+        ByteArrayOutputStream stdout = new ByteArrayOutputStream();
+        ByteArrayOutputStream stderr = new ByteArrayOutputStream();
```
jaikiran: Hi Craig, like Stefan already said, this patch is a really good addition to Ant. Thank you for that.

The only question/suggestion I have is, should we instead just pass `System.out` and `System.err` instead of creating a `ByteArrayOutputStream`? That way the actual output and errors are logged instead of we deciding to print it only when the tool execution fails.

craigpell: That was what I tried first, but the output and error were never showing up at all in unit tests.  (Also, my observation has been that the tool produces no output when it runs successfully.)

bodewig:
BSD is perfectly fine, thanks.

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Windows/103/

asfgit:
https://builds.apache.org/job/Ant%20Github-PR-Linux/97/

src/main/org/apache/tools/ant/taskdefs/modules/Jmod.java
bodewig: Test setup captures streams so it may become difficult. Would using `LogOutputStreams` do which would send the output directly to An't logging system?

bodewig:
Many thanks again, @craigpell . I've merged this but would like to resolve the discussion about the log output.

With b74d5b366 I've committed a branch (logoutput-jmod-link) where I use `LogOutputStream` in `link` - I'm not sure why one would suppress the output at all, but I've never worked with the tool. A similar approach (or one that does away with the `ByteArrayOutputStream`s completely and removes the output from the exception and simply channels the output to Ant's log unconditionally) would work for `jmod` as well.

/cc @jaikiran

craigpell:
I restored the LogOutputStream usage and confirmed that the output is not visible in unit test results.  It's not even captured in the system-out and system-err elements of the TEST-*.xml files.

I remember spending a good amount of time trying to find a method that would make the log show up, but I was never able to do it.

jaikiran:
>> I restored the LogOutputStream usage and confirmed that the output is not visible in unit test results. It's not even captured in the system-out and system-err elements of the TEST-*.xml files.

Hi @craigpell, the test(s) use the `BuildFileRule`[1] which as Stefan mentioned, redirect the output/error streams to `StringBuffer`(s)[2] which are stored in memory. Those logs are available (for tests) through the `getFullLog` and other such methods on the `BuildFileRule`.



[1] https://github.com/apache/ant/pull/80/files#diff-1bb76afc7ce4a535b0c580f9d040aff9R68
[2] https://github.com/apache/ant/blob/master/src/tests/junit/org/apache/tools/ant/BuildFileRule.java#L175
[3] https://github.com/apache/ant/blob/master/src/tests/junit/org/apache/tools/ant/BuildFileRule.java#L106

PR #82, opened 2018-12-20
Use valid markup

twogee:
For 100% compliance, all HTML documents in the manual should add `<!DOCTYPE html>` and replace `<meta http-equiv="Content-Language" content="en-us">` with `<html lang="en-us">`


bodewig:
Thanks!

PR #100, opened 2019-09-03
Send FQDN instead of simple hostname in SMTP HELO command

mguessan:
Some SMTP servers block messages when HELO command contains a simple hostname instead of a fully qualified domain name.

This path fixes this by using getCanonicalHostName instead of getHostName

### DIFF
--- a/check.xml
+++ b/check.xml
@@ -105,7 +105,7 @@
 
   <!-- Apache Rat: Release Auditing Tool -->
   <property name="rat.version"
-            value="0.12"
+            value="0.13"
             description="Which version of Apache Rat to use"/>
   <property name="rat.report.dir"
             value="${reports.dir}/rat"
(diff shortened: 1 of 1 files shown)


bodewig:
this is ok to test

jaikiran:
The test failures being reported in those runs, for `org.apache.tools.mail.MailMessageTest` look related. The tests expect a different host name which no longer matches the one being sent. 
I haven't yet had a look at the existing code/tests, but we will have to review if this backward incompatibility is OK or if we should make this configurable and default to the previous behaviour.

mguessan:
MailMessageTest is now fixed, however there are still 4 test failures. They don't seem related to this PR.

twogee:
Looking at the specs, shouldn't the task be made [RFC-5321](https://tools.ietf.org/html/rfc5321#page-93)-compliant?  For unless I am mistaken, [RFC-821](https://tools.ietf.org/html/rfc821#page-27) says HELO must use _domain_...

twogee:
Looking further, RFC-821 states "The argument field contains the host name of the sender-SMTP." and the formal syntax suggests that "domain" means FQDN. RFC-2821 clarifies
```
  The argument field contains the fully-qualified domain name
   of the SMTP client if one is available.  In situations in which the
   SMTP client system does not have a meaningful domain name (e.g., when
   its address is dynamically allocated and no reverse mapping record is
   available), the client SHOULD send an address literal (see section
   4.1.3), optionally followed by information that will help to identify
   the client system. 
```
RFC-5321 deprecates the use of address literals.

mguessan:
Ok, so basically getCanonicalHostName returns exactly what we need, right ?

twogee:
Yes, and we should change HELO to EHLO.

mguessan:
Sure, not very comfortable with git yet... 

I will probably have to create a new PR as this one is linked to master on my side (my mistake), right ?

PR #101, opened 2019-09-13
Send FQDN instead of simple hostname in SMTP HELO command

mguessan:
Some SMTP servers block messages when HELO command contains a simple hostname instead of a fully qualified domain name.

This path fixes this by using getCanonicalHostName instead of getHostName
See https://github.com/apache/ant/pull/100


jaikiran:
Thank you @mguessan for this PR. I have merged this and added you in our contributors list.

PR #103, opened 2019-10-21
Made some variable names more consistent with the other parts.

euske:
Hello, we're developing an automated system that detects inconsistent variable names in a large software project. Our system checks if each variable name is consistent with the other variables in its usage, and proposes correct candidates if inconsistency is found. This is a part of academic research that we hope to publish soon, but as a part of the evaluation, we applied our systems to your projects and got a few interesting results. We carefully reviewed our system output and manually created a patch to correct a few variable names. We would be delighted if this patch is found to be useful. If you have a question or suggestion regarding this patch, we'd happily answer. Thank you.

P.S. our patch is purely for readability purposes and does not change any functionality. A couple of issues that we've noticed were left untouched. For example, there are some overlap use between "msgLevel" "logLevel" and "priority" for logging, but we weren't sure whether they actually refer to the same concept or not, so we didn't change them.

### DIFF
--- a/src/main/org/apache/tools/ant/BuildException.java
+++ b/src/main/org/apache/tools/ant/BuildException.java
@@ -74,15 +74,15 @@ public class BuildException extends RuntimeException {
      * Constructs an exception with the given message and exception as
      * a root cause and a location in a file.
      *
-     * @param msg A description of or information about the exception.
-     *            Should not be <code>null</code> unless a cause is specified.
+     * @param message A description of or information about the exception.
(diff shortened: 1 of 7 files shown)


janmaterne:
It seems ok to me. As far as I could see, only method local variables are renamed. So we don't have any clue with BWC.

jaikiran:
this is ok to test

janmaterne:
merged
Thanks for the contribution.

PR #98, opened 2019-06-14
Use commandline argument file for all options, but -J, for case javac…

slobokot:
Currently only java files are passed using commandline argument file for javac.
There are times when classpath is longer than 8000 characters. on unix that works, but on windows it doesn't, see https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation. That leads to exception like
```
at org.apache.tools.ant.taskdefs.compilers.DefaultCompilerAdapter.executeExternalCompile(DefaultCompilerAdapter.java:601)
        at org.apache.tools.ant.taskdefs.compilers.JavacExternal.execute(JavacExternal.java:62)
        at org.apache.tools.ant.taskdefs.Javac.compile(Javac.java:1395)
        at org.apache.tools.ant.taskdefs.Javac.execute(Javac.java:1121)
        ...
Caused by: java.io.IOException: Cannot run program "c:\...\javac.exe" (in directory "..."): CreateProcess error=206, The filename or extension is too long
        at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
        at java.lang.Runtime.exec(Runtime.java:620)
        at org.apache.tools.ant.taskdefs.launcher.Java13CommandLauncher.exec(Java13CommandLauncher.java:58)
        at org.apache.tools.ant.taskdefs.Execute.launch(Execute.java:426)
        at org.apache.tools.ant.taskdefs.Execute.execute(Execute.java:440)
        at org.apache.tools.ant.taskdefs.compilers.DefaultCompilerAdapter.executeExternalCompile(DefaultCompilerAdapter.java:598)
        ... 33 more
Caused by: java.io.IOException: CreateProcess error=206, The filename or extension is too long
        at java.lang.ProcessImpl.create(Native Method)
        at java.lang.ProcessImpl.<init>(ProcessImpl.java:386)
        at java.lang.ProcessImpl.start(ProcessImpl.java:137)
        at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
        ... 38 more
```

This pull requests puts all but -J options into commandline argument file. -J options can't be specified in commandline argument file as per documentation: https://docs.oracle.com/en/java/javase/11/tools/javac.html

PR #105, opened 2019-10-23
Fix regression introduced by commit "Use commandline argument file fo…

slobokot:
Following change https://github.com/apache/ant/pull/98 caused regression in a case when compiler arg is argfile, for example
```
<javac fork="true">
    <compilerarg value="@path-to-args-file"/>
</javac>
```
This change moves '@argfiles' options to the beginning of commandLine arguments, same as -J options, so that all other options are written to file. Tested on windows with 5000 characters long command line for javac.

Related bug reports: 
https://bugs.archlinux.org/task/64055
https://bugs.openjdk.java.net/browse/JDK-8231316


jaikiran:
this is ok to test

jaikiran:
@slobokot Thank you for this patch. Looks good to me. This has been merged upstream.

PR #106, opened 2019-10-24
Support SSH config file in the ssh tasks

ebourg:
Here is a PR improving the sshexec, sshsession and scp tasks to support the SSH config file (typically ~/.ssh/config). This file defines the username and key file to be used per host. If Ant could support it that would avoid duplicating the key information from the SSH config in the Ant builds.

Ivy already supports it for its SSH resolver: https://ant.apache.org/ivy/history/latest-milestone/resolver/ssh.html

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/optional/ssh/SSHBase.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/ssh/SSHBase.java
@@ -18,6 +18,11 @@
 
 package org.apache.tools.ant.taskdefs.optional.ssh;
 
+import java.io.File;
+import java.io.IOException;
+
+import com.jcraft.jsch.ConfigRepository;
+import com.jcraft.jsch.OpenSSHConfig;
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Project;
 import org.apache.tools.ant.Task;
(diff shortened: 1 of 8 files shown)


src/main/org/apache/tools/ant/taskdefs/optional/ssh/SSHBase.java
```diff
@@ -235,6 +259,37 @@ public void init() throws BuildException {
...
+            ConfigRepository.Config config = null;
+            try {
+                config = OpenSSHConfig.parseFile(sshConfig).getConfig(host);
+            } catch (IOException e) {
+                throw new BuildException("Failed to load the SSH configuration file " + sshConfig, e);
+            }
+            
+            host = config.getHostname();
+            
+            if (userInfo.getName() == null) {
```
jaikiran: Should this check perhaps happen before even trying to load the `sshConfig` file? So as to avoid the whole loading and parsing of the file?

Perhaps load the file only if userInfo doesn't have the name or keyfile available?

jaikiran:
@ebourg This looks good to me. Thank you for providing this patch - looks really useful. I just have relatively minor comment that I have added to this PR. Once we decide on that, I'll go ahead and merge this.

ebourg:
Thank you for the quick review. I've pushed a change addressing your comment.

jaikiran:
Thank you. This is now merged.

ebourg:
Thanks!

PR #110, opened 2019-11-05
Backport magic names

twogee:


jaikiran:
What's the context behind proposing these changes to 1.9.x branch?

twogee:
The rationale for revision of MagicNames/MagicTestNames is threefold:

- make the use of special property names/constants explicit;
- establish a single definition of property names/constants (there's a duplication between MagicNames and ProjectHelper; a rework may be necessary if MagicNames should be for property names and ProjectHelper for constants only);
- BuildFileTest/BuildFileRule should ignore basedir set by Maven when JUnit tests are run by Surefire (like it or not, developers of Ant plugins are using it).

jaikiran:
1.9.x branch is meant only for very limited changes. We don't have any plans to introduce changes like these in that branch.

twogee:
I'd like to keep the PR open and see if it generates more interest, if you don't mind.

jaikiran:
I'm not going to close this and will let it stay open. 

However, this is just a waste of reviewers' time to create such PRs when it's clearly known that 1.9.x branch and refactoring changes like these in general, aren't accepted. It's not the first time this has been clearly explained over the years and just repeating this over and over again in the hopes of pushing these in, is just complete disregard of time spent by others in engaging in such discussions.

This PR gets a official -1 from me.

twogee:
I believe that the proposed changes have a merit, and I split the commit to reduce the scope.

PR #111, opened 2019-11-06
build.sysclasspath is a magic name

twogee:


twogee:
magicAntProperty test seems to fail on Windows all the time, perhaps it should be made U*X-only?

PR #112, opened 2019-11-08
Update AntClassLoader.java

majkelBB:
Shouldn't usage of pathComponents be threadsafe here and in other places where code iterates over pathComponents?


jaikiran:
this is ok to test

bodewig:
The way `AntClassLoader` wants to be used (and is used inside of Ant's codebase) is a single-threaded setup phase followed by potentially multithreaded use of if where its setup is no longer changed.

Personally I'd rather document this than try to sprinkle in a few `synchronized`s.

PR #113, opened 2019-12-07
Avoid illegal control characters in HTML

adamretter:
Previously it was possible to create illegal HTML, by copying control characters from the XML. This fixes that by escaping them.


jaikiran:
this is ok to test

adamretter:
Can anyone tell me if this test case failure is related to my change? I do not see how it can be.

jaikiran:
Hello @adamretter, those tests failures aren't related to this change. I haven't fully reviewed the change, if someone else reviews it before me, they will merge it. Else I'll take a look later tonight or tomorrow.

PR #122, opened 2020-03-03
Bugfix/catch dynamic test before each

mguillem:
Ensure that every failed test is reported by the `LegacyXmlResultFormatter`. Currently failures in `@BeforeEach` with a `@TestFactory` are not visible in the XMLreports. Possibly not fully correct `classname` and `name` attributes are far better than an invisible failure.

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/LegacyXmlResultFormatter.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/LegacyXmlResultFormatter.java
@@ -263,13 +263,13 @@ class LegacyXmlResultFormatter extends AbstractJUnitResultFormatter implements T
             }
             
             if (!parentClassSource.isPresent()) {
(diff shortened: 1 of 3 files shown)


mguillem:
Hi Ant Team. Can someone provide feedback to this PR? Thanks.

jaikiran:
Hello @mguillem, sorry I just saw this now. Is there a junitreport output that you can show with and without this change? If not, I can test/check this myself in a few days and see how this change then impacts the reports.

jaikiran:
this is ok to test

mguillem:
@jaikiran do you still need something from me? Failed test seems unrelated to this PR.

jaikiran:
Hello Marc, sorry this took this long. I saw this PR again this week and what you raise here is a valid bug. I've pushed a commit which should address this and should be available in our upcoming Ant 1.10.14 release.

Please continue to raise any other issues you might see in this task or Ant in general. Sorry again about the delay.

PR #125, opened 2020-04-14
junit4 Parameterized not reported

TheGoesen:
The LegacyXmlResultsFormatter does not report induvidual testcases for Paramerized tests.
This affects all Junit4 Tests annotated with.

@RunWith(Parameterized.class)

Maybe it would also be better if the code if the code was rewritten that in case no ClassSource was found some other string would be written instead of just silently skipping the testcase.


jaikiran:
Hello @TheGoesen, is this still an issue in the recently released 1.10.9 of Ant? If so, would you be able to create a bugzilla issue[1] for this with a sample build.xml to quickly reproduce this issue? 
[1] https://bz.apache.org/bugzilla/enter_bug.cgi?product=Ant

TheGoesen:
I opened https://bz.apache.org/bugzilla/show_bug.cgi?id=64952

jaikiran:
Hello @TheGoesen, thank you for this PR. I've merged this and closed the linked bugzilla issue. I've also added you (as Gösen) to our contributors list https://github.com/apache/ant/blob/master/CONTRIBUTORS#L161. If you want me to name you different in that list do let me know.

PR #127, opened 2020-05-23
fix two typos in rmic manual page

esaulpaugh:


bodewig:
Thank you @esaulpaugh

PR #129, opened 2020-06-21
Extract LegacyPlainResultFormatter.Stats#appendElapsed

simon04:


bodewig:
Many thanks. By which name would you want us to credit you in Ant's CONTRIBUTORS file @simon04 ?

simon04:
"Simon Legner" is fine, thank you! :)

bodewig:
Thanks again.

PR #130, opened 2020-07-27
Correct javadoc

twogee:


bodewig:
I don't understand why you want to change the filelist examples to use fileset.

The other file is a Java file used inside of a test for rmic.

twogee:
Using `<filelist/>` throws an error: `filelist doesn't support the nested "modified" element.`, see Bugzilla report.

bodewig:
For reference: https://bz.apache.org/bugzilla/show_bug.cgi?id=64612

I was under the false impression we had added selectors to all resource collections, obviously I am wrong. Sorry about that.

PR #131, opened 2020-07-27
Another getCheckedRef() refactoring

twogee:


bodewig:
I don't see any reason to deprecate `DATA_TYPE_NAME`, You don't need to do that for the refactoring aynway.

PR #133, opened 2020-07-28
Class.newInstance() is deprecated in Java 9+

twogee:


src/main/org/apache/tools/zip/ExtraFieldUtils.java
```diff
@@ -81,10 +87,10 @@ public static void register(Class<?> c) {
      * @since 1.1
      */
     public static ZipExtraField createExtraField(ZipShort headerId)
-        throws InstantiationException, IllegalAccessException {
+            throws InstantiationException, IllegalAccessException, NoSuchMethodException, InvocationTargetException {
```
bodewig, line 90: this changes the signature of a public method, I'm afraid we can not do this without breaking backwards compatibility.

twogee: thanks, missed that, wrapped in InstantiationException.

PR #138, opened 2020-10-04
Update the optional dependencies

twogee:


bodewig:
@twogee the switch to jakarta.mail is not backwards compatible. I don't think we can do it that way at all. Rather we'll need to introduce a new type of Mailer and an additional jar with different dependencies, I'm afraid.

bodewig:
Thank you @twogee - do you think you could split the Jakarta Mailer support changes from the other changes?

The changes to `ClasspathUtils` seem to be completely unrelated to either and the chunk that removes string concatenation is completely unnecessary anyway.

twogee:
Thanks, @bodewig, and sorry about missing the extraneous changes.
I will open a new PR for changes re Jakarta Mail specifically.
The change to `ClasspathUtils` is related to an attempt to avoid nested `try` in `EmailTask#getMailerImplementation()`.

PR #140, opened 2020-10-28
#64854 Add preserveduplicates option to ResourceList

TheConstructor:
I propose this to solve https://bz.apache.org/bugzilla/show_bug.cgi?id=64854

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -285,6 +285,7 @@ Matthew Watson
 Matthew Yanos
 Matthias Bhend
 Matthias Gutheil
+Matthias Johann Vill
 Michael Bayne
 Michael Clarke
 Michael Davey
--- a/contributors.xml
+++ b/contributors.xml
@@ -1184,6 +1184,11 @@
     <last>Gutheil</last>
   </name>
   <name>
+    <first>Matthias</first>
+    <middle>Johann</middle>
+    <last>Vill</last>
+  </name>
+  <name>
     <first>Michael</first>
     <last>Bayne</last>
   </name>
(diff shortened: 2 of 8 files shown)


TheConstructor:
I just send the e-mail with the ICLA. Feel free to improve my code.

bodewig:
You removed the `volatile` attribute `cached´ from `ResourceList`. Th way code is structured now there is a race-condition where one thread my be using a cache that hasn't been initialized completely. I'm afraid you need to bring back `cached` redundant as it may seem.

Apart from that I've only got a few nits:

* I'm not sure about the name `WriteableResourceCollection` - `AppendableResourceCollection` maybe?
* please add since markers with 1.10.10 to new classes/methods and to the manual
* more importantly, add yourself to CONTRIBUTORS (plain text) and contributors.xml. The files are supposed to be ordered lexicographically by first name.

Many thanks.

TheConstructor:
I checked that `cachedResources` is only ever accessed through `cache()`, which is `synchronized`. Is this really a problem? Would declaring `cachedResources` `volatile` and/or assigning the field just after populating the cache help in your opinion?

To your nits:
- I am happy to change the name
- I wasn't quite sure which version could be next, so I thought it would be good to have something in place, where I can pick up work. -> will insert the version
- Ok

bodewig:
I completely overlooked `cache` was synchronized. Since all read an write operations on `cachedResources` are synchronized, I don't think a problem exists there. Sorry.

TheConstructor:
@bodewig Thank you again for looking into this and my other PR!

PR #141, opened 2020-10-29
#64855 Add filterbeforeconcat option to Concat

TheConstructor:
I propose this to solve https://bz.apache.org/bugzilla/show_bug.cgi?id=64855

To implement optional filtering before concatenation I moved the end-of-file-linebreak-fix into it's own LastLineFixingReader so that MultiReader.nextReader() can apply line-break-fixing and/or filtering on demand.

### DIFF
--- a/manual/Tasks/concat.html
+++ b/manual/Tasks/concat.html
@@ -125,6 +125,15 @@
         <td>No; default is <q>false</q></td>
       </tr>
       <tr>
+        <td>filterbeforeconcat</td>
+        <td>If this attribute is set to <q>true</q>, the task applies the filterchain to each
+          input after applying <code>fixlastline</code>. If this attribute is <q>false</q>, concat
+          will apply the filterchain only once to the already concatenated inputs. Filtering of
+          <code>header</code> and <code>footer</code> is not affected by this setting.
+          <em>Since Ant 1.10.10</em></td>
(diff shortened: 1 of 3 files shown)


TheConstructor:
Is there anything I can/need to do?

bodewig:
@TheConstructor no, you've done all there is to do, many thanks. The silence is totally out fault, we just seem all to be busy right now. I'll find some time looking into this and #140 this week.

TheConstructor:
@bodewig thank you for the heads-up and for looking into this!

bodewig:
I like the approach of moving the logic for fixing the lat lines to a separate filter. Kudos for adding tests ;-)

A different approach could be to allow an alternative set of filters to be applied before concatenating, i.e. allowing users to define two sets of filters one to apply on each input stream individually and one to be applied to the merged stream. Naming could become an issue (and I've got a long track of picking bad names myself).

I'm not asking you to change your PR, I'd rather like to discuss the different approaches to see which would be best. So what do you think?

TheConstructor:
I briefly thought about that, but thought it would at another layer of complexity. In the end you could wrap a <concat filterbeforeconcat="true"> into another <concat> without `filterbeforeconcat`, if you want to filter before and after concatenation.

bodewig:
Yes, you are right.

PR #142, opened 2020-12-10
Tar-task: using <tarfileset> should preserve symlinks

mstrap:


bodewig:
many thanks @mstrap 

I'd like to add you to the contributors list (CONTRIBUTORS and contributors.xml), which name shall I use?

mstrap:
Thanks, @bodewig ! Please use "Marc Strapetz".

bodewig:
added, thanks @mstrap

PR #145, opened 2021-03-13
Override ZipOutputStream.write(int)

helfper:
`java.util.zip.ZipOutputStream` extends from `java.util.zip.DeflaterOutputStream` which overrides `void write(int b)` to redirect its calls to `void write(byte[] b, int off, int len)`.

`org.apache.tools.zip.ZipOutputStream` doesn't extend the same class and currently doesn't override `void write(int b)`. This means that calls to this method end up in `FilterOutputStream.write(int b)`, which in turn passes them through to the underlying `OutputStream`. This bypasses all the logic in `ZipOutputStream.write(byte[] b, int offset, int len)` (deflation and whatnot) which produces corrupted ZIP files.

One example where the method `void write(int b)` is ultimately invoked is by calling `java.util.jar.Manifest.write(zipOutputStream)`.

This PR fixes this by overriding the method `void write(int b)` in `org.apache.tools.zip.ZipOutputStream` in the same way as `java.util.zip.DeflaterOutputStream` does.

### DIFF
--- a/src/main/org/apache/tools/zip/ZipOutputStream.java
+++ b/src/main/org/apache/tools/zip/ZipOutputStream.java
@@ -326,6 +326,11 @@ public class ZipOutputStream extends FilterOutputStream {
     private final Calendar calendarInstance = Calendar.getInstance();
 
     /**
+     * Temporary buffer used for the {@link #write(int)} method.
+     */
+    private final byte[] oneByte = new byte[1];
+
+    /**
      * Creates a new ZIP OutputStream filtering the underlying stream.
      * @param out the outputstream to zip
      * @since 1.1
@@ -899,6 +904,19 @@ public class ZipOutputStream extends FilterOutputStream {
      */
     public boolean canWriteEntryData(ZipEntry ae) {
         return ZipUtil.canHandleEntryData(ae);
+    }
+
+    /**
+     * Writes a byte to ZIP entry.
+     *
+     * @param b the byte to write
+     * @throws IOException on error
+     * @since Ant 1.10.10
+     */
+    @Override
+    public void write(int b) throws IOException {
+        oneByte[0] = (byte) (b & 0xff);
+        write(oneByte, 0, 1);
     }
 
     /**
--- a/src/tests/junit/org/apache/tools/zip/ZipOutputStreamTest.java
+++ b/src/tests/junit/org/apache/tools/zip/ZipOutputStreamTest.java
@@ -21,8 +21,14 @@ package org.apache.tools.zip;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.jar.JarFile;
+import java.util.jar.Manifest;
+import java.util.zip.ZipInputStream;
 
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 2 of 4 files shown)


src/main/org/apache/tools/zip/ZipOutputStream.java
```diff
@@ -901,6 +901,20 @@ public boolean canWriteEntryData(ZipEntry ae) {
...
+    /**
+     * Writes a byte to ZIP entry.
+     *
+     * @param b the byte to write
+     * @throws IOException on error
+     * @since Ant 1.10.10
+     */
+    @Override
+    public void write(int b) throws IOException {
+        byte[] buf = new byte[1];
```
bodewig, line 913: We could allocate a one-byte array and reuse that for all calls. Of course then we'd need to think about thread safety issues. So maybe that's not wort the trouble given that so far nobody seems to have used that method anyway :-)

src/tests/junit/org/apache/tools/zip/ZipOutputStreamTest.java
```diff
@@ -70,4 +76,17 @@ public void testAdjustToLong() {
...
+        ZipOutputStream zos = new ZipOutputStream(baos);
+        zos.putNextEntry(new ZipEntry(JarFile.MANIFEST_NAME));
+        new Manifest().write(zos);
+        zos.close();
+        ByteArrayInputStream bais = new ByteArrayInputStream(baos.toByteArray());
+        ZipInputStream zis = new ZipInputStream(bais);
+        zis.getNextEntry();
+        zis.closeEntry();
+        zis.close();
+    }
```
bodewig, line 91: please use try-with-resources for all streams even though we are only writing to memory.

bodewig:
Many thanks - interesting this has been there for more then a decade. Most likely this is because Ant is supposed to be the only consumer of the class and doesn't use the single byte write method itself. For general use I'd recommend to use `ZipArchiveOutputStream` of Apache Commons Compress which is a grandchild of Ant's zip package.

@helfper please add yourself to the contributors.xml and CONTRIBUTORS files in Ant's top level.

src/main/org/apache/tools/zip/ZipOutputStream.java
helfper: Agreed, I thought about that but was wondering why `DeflaterOutputStream` didn't do it. I did it now as this class is not thread-safe anyway, and this allocation on every write may keep the GC unnecessarily busy. I checked Apache Commons Compress since you mentioned it and it's also it's done there: https://github.com/apache/commons-compress/blob/master/src/main/java/org/apache/commons/compress/archivers/ArchiveOutputStream.java#L52

src/tests/junit/org/apache/tools/zip/ZipOutputStreamTest.java
helfper: Done now.

helfper:
@bodewig I've updated the contributors files now.

I've found the issue in https://github.com/johnrengelman/shadow which uses Ant's `ZipOutputStream`. Out of curiosity, why doesn't Ant use Commons Compress's `ZipArchiveOutputStream`?

bodewig:
Ant predates Commons Compress by several years.

Actually Ant's core doesn't have any dependencies at all, this has always been the case. Initially the XML parser had to be provided before JAXP became part of the standard class library. This is an important goal of Ant. We can bootstrap Ant with just a JDK and nothing else.

PR #146, opened 2021-03-18
correct a typo in the install part of manual (file separator)

chenyuheng:
File separator in Windows is backslash instead of slash.


bodewig:
Thank you @chenyuheng

PR #147, opened 2021-04-30
Make Watchdog for JUnitLauncherTask customizable.

alex-ninja:
### Context
Currently I'm working on migration of Cassandra to JUnit5 (https://issues.apache.org/jira/browse/CASSANDRA-16630). There is a nice customization that is used there: before terminating a test by timeout, we get all stack traces via `jstack` command. It is very useful for troubleshooting purposes. You can check the reference implementation for JUnit4: https://github.com/apache/cassandra/blob/trunk/test/unit/org/apache/cassandra/JStackJUnitTask.java. Unfortunately it is impossible to do the same customization for `JUnitLauncherTask` (unlikely to `JUnitTask`) because everything is hidden (private) now. We want to have a way to easily extend `Watchdog` created for a task/test monitoring.

### Summary of the changes
1. Moved Watchdog creation into a separate method
2. Made the method protected


ddevienne:
Looks perfectly reasonable to me, FWIW. +1.

PR #148, opened 2021-05-19
Fix echoxml example in manual

babasaikiran:
The example of the [echoxml](http://ant.apache.org/manual/Tasks/echoxml.html) task seems to be incorrect.


babasaikiran:
sorry, i didn't realize the example hadn't to be part of project and a target.

PR #149, opened 2021-07-11
Ant doesn't call the superclass constructor and pass in the specified parent classloader

basil:
See [JENKINS-21579](https://issues.jenkins.io/browse/JENKINS-21579) and [JENKINS-22310](https://issues.jenkins.io/browse/JENKINS-22310). Back in 2014, Kohsuke Kawaguchi (the creator of Jenkins) forked `AntClassLoader` from Ant 1.8.3 into Jenkins core to apply [the following fix](https://github.com/jenkinsci/jenkins/commit/9a2882dd70) to it:

```
commit 9a2882dd70 (tag: changes/76)
Author: Kohsuke Kawaguchi <kk@kohsuke.org>
Date:   Mon Mar 24 23:04:54 2014

    [FIXED JENKINS-21579] UberClassLoader.findResources() improvements
    
    This is just a locally patched version, so not meant to be used outside
    core.
```

I am now looking into unforking `AntClassLoader` in Jenkins core so that we can reuse the latest upstream version from Ant 1.10.x. As far as I can tell the fix remains valid and should be contributed upstream.

[This comment](https://issues.jenkins.io/browse/JENKINS-22310?focusedCommentId=197405&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-197405) specifically contains Kohsuke's analysis describing the cause of the problem and the motivation for the fix.

This fix has been battle tested on every Jenkins installation for the past 7 years.

### DIFF
--- a/src/main/org/apache/tools/ant/AntClassLoader.java
+++ b/src/main/org/apache/tools/ant/AntClassLoader.java
@@ -273,6 +273,7 @@ public class AntClassLoader extends ClassLoader implements SubBuildListener, Clo
      * @param classpath The classpath to use to load classes.
      */
     public AntClassLoader(final ClassLoader parent, final Project project, final Path classpath) {
+        super(parent);
         setParent(parent);
         setClassPath(classpath);
         setProject(project);
@@ -297,6 +298,7 @@ public class AntClassLoader extends ClassLoader implements SubBuildListener, Clo
      *                elements are set up to start with.
      */
     public AntClassLoader(final Project project, final Path classpath) {
+        super(parent);
         setParent(null);
         setProject(project);
         setClassPath(classpath);
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -43,6 +43,7 @@ Avik Sengupta
 Balazs Fejes 2
 barney2k7
 Bart Vanhaute
+Basil Crow
 Ben Galbraith
 Ben Gertzfield
 Benjamin Burgess
--- a/contributors.xml
+++ b/contributors.xml
@@ -200,6 +200,10 @@
     <last>Vanhaute</last>
   </name>
   <name>
+    <first>Basil</first>
+    <last>Crow</last>
+  </name>
+  <name>
     <first>Benjamin</first>
     <last>Burgess</last>
   </name>


jaikiran:
Hello @basil, although this change looks straightforward and seems to have been in-use in Jenkins for 7 odd years, I will need some inputs from others before merging this change.

More specifically, I had a look at the `AntClassLoader`'s constructor (the one being changed) and noticed that it has code that deals with the `parent` in a very specific way. Plus, the commit history of that class suggests that there was some thought given around how the `parent` is currently used/set. I don't have that background knowledge, except that it somehow relates to https://bz.apache.org/bugzilla/show_bug.cgi?id=35436. Perhaps @bodewig or someone more knowledgeable with this code, might have inputs on whether or not to merge this change.

src/main/org/apache/tools/ant/AntClassLoader.java
```diff
@@ -297,6 +298,7 @@ public AntClassLoader() {
      *                elements are set up to start with.
      */
     public AntClassLoader(final Project project, final Path classpath) {
+        super(parent);
```
jaikiran, line 301: I think to make it clear what's being done here, it's better to do `super(null)` instead of using `parent` which is `null`.

basil:
Thanks for the quick reply @jaikiran! The bug you linked to is very helpful. I am pleased to report that the original script given by Kohsuke in [this comment](https://issues.jenkins.io/browse/JENKINS-22310?focusedCommentId=197405&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-197405) still reproduces the problem in Jenkins core when the `super(parent)` calls are removed. So I have a minimal reproducible example that can be used to validate different approaches.

I now see that Kohsuke's approach was Solution 1 described in [Bug 35436](https://bz.apache.org/bugzilla/show_bug.cgi?id=35436) by Rainer Noack:

> Remove the local property and it's setter-method. Allow setting of the parent only in the constructors. Use `ClassLoader`'s parent property. Advantages: IMHO, the cleanest solution. Disadvantages: Many build brakes. Fazit: Not usable

The latest comment by @bodewig states:

> I don't think we would achieve much by setting `ClassLoader.parent` […]. In order to close this issue I've simply added a `getConfiguredParent` method in svn revision 796649

This was Solution 2 described in the original bug by Rainer Noack:

> Provide a "special" getter like `getRealParent()` Advantages: works with existing environments. Disadvantages: such a shame Fazit: No better way !?

So Stefan decided to go in a different, but equally valid, direction from Kohsuke. `AntClassLoader` has a mutable `parent` field, but its (no pun intended) parent `ClassLoader` class has an immutable `parent` field that can only be set by the constructor. The `getParent()` method is `final`, so `AntClassLoader` can't override it to return `AntClassLoader`'s mutable `parent` field. So @bodewig added a new `getConfiguredParent()` method which returns `AntClassLoader`'s mutable `parent` field. This makes sense, but the problem is that this change was incomplete. Other code still calls `getParent()` in several places. In particular, consider the code Kohsuke referred to in his analysis in JENKINS-22310:

```java
if (parent != null && (!parentHasBeenSearched || parent != getParent())) {
    // Delegate to the parent:
    base = parent.getResources(name);
    // Note: could cause overlaps in case
    // ClassLoader.this.parent has matches and
    // parentHasBeenSearched is true
} else {
    // ClassLoader.this.parent is already delegated to for example from
    // ClassLoader.getResources, no need:
    base = Collections.emptyEnumeration();
}
```

This calls `getParent()`, which will return the immutable `parent` field from the parent `ClassLoader` class rather than `AntClassLoader`'s mutable `parent` field. This is clearly a bug in `AntClassLoader`. This code should be using `getConfiguredParent()`. Sure enough, with this change I can no longer reproduce the problem in Jenkins:

```java
@@ -961,7 +967,7 @@ public class AntClassLoader extends ClassLoader implements SubBuildListener, Clo
         throws IOException {
         final Enumeration<URL> mine = new ResourceEnumeration(name);
         Enumeration<URL> base;
-        if (parent != null && (!parentHasBeenSearched || parent != getParent())) {
+        if (parent != null && (!parentHasBeenSearched || parent != getConfiguredParent())) {
             // Delegate to the parent:
             base = parent.getResources(name);
             // Note: could cause overlaps in case
```

There's still another problematic use of `getParent()` in `getRootLoader()`. This doesn't appear to be a problem for Jenkins, but it is also technically incorrect. `getRootLoader()` recursively walks the `ClassLoader` hierarchy via `getParent()` until it reaches the root node. But if any of those `ClassLoader`s happen to be an `AntClassLoader` with a broken `getParent()` method, that won't work. The solution is to treat `AntClassLoader` specially:

```java
@@ -841,12 +839,20 @@ public class AntClassLoader extends ClassLoader implements SubBuildListener, Clo
      */
     private ClassLoader getRootLoader() {
         ClassLoader ret = getClass().getClassLoader();
-        while (ret != null && ret.getParent() != null) {
-            ret = ret.getParent();
+        while (ret != null && getParent(ret) != null) {
+            ret = getParent(ret);
         }
         return ret;
     }
 
+    private static ClassLoader getParent(ClassLoader cl) {
+        if (cl instanceof AntClassLoader) {
+            return ((AntClassLoader) cl).getConfiguredParent();
+        } else {
+            return cl.getParent();
+        }
+    }
+
```

I tested this change with Jenkins, and Jenkins seems to work fine with it (although it was never broken in the first place).

In my opinion, @bodewig's change in 2009 that introduced `getConfiguredParent()` should have updated both of these callers.

It's not to late to do it now in 2021.

basil:
I separated out my findings into two new PRs for us to discuss further. #152 covers the issue with walking the class loader hierarchy. #151 covers the problem with `findResources(String, boolean)` - which is a little more complicated than it seems. Basically, the existing implementation no longer makes sense after commit 17527b6, and updating it to take commit 17527b6 into account also fixes the Jenkins bug.

basil:
Closing in favor of #151.

PR #151, opened 2021-07-21
findResources(String, boolean) can unnecessarily search the parent

basil:
See [#149 (comment)](https://github.com/apache/ant/pull/149#issuecomment-884318912). I verified that without this fix, I could reproduce the original problem described by Kohsuke Kawaguchi in [JENKINS-22310 (comment)](https://issues.jenkins.io/browse/JENKINS-22310?focusedCommentId=197405&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-197405). I verified that with this fix, I could no longer reproduce the original problem.

### DIFF
--- a/src/main/org/apache/tools/ant/AntClassLoader.java
+++ b/src/main/org/apache/tools/ant/AntClassLoader.java
@@ -936,9 +936,6 @@ public class AntClassLoader extends ClassLoader implements SubBuildListener, Clo
      * data (images, audio, text, etc) that can be accessed by class
      * code in a way that is independent of the location of the code.
      *
-     * <p>Would override getResources if that wasn't final in Java
-     * 1.4.</p>
-     *
      * @param name name of the resource
      * @return possible URLs as enumeration
      * @throws IOException if something goes wrong
@@ -982,26 +979,21 @@ public class AntClassLoader extends ClassLoader implements SubBuildListener, Clo
      *
      * @param name The resource name to search for.
      *             Must not be <code>null</code>.
-     * @param parentHasBeenSearched whether ClassLoader.this.parent
-     * has been searched - will be true if the method is (indirectly)
-     * called from ClassLoader.getResources
+     * @param skipParent whether to skip searching the parent first - will be false if the method is
+     *     invoked from {@link #getResources(String)} or {@link #getNamedResources(String)} and true
+     *     if the method is invoked from {@link #findResources(String)}.
      * @return an enumeration of URLs for the resources
      * @exception IOException if I/O errors occurs (can't happen)
      */
     protected Enumeration<URL> findResources(final String name,
-                                             final boolean parentHasBeenSearched)
+                                             final boolean skipParent)
         throws IOException {
         final Enumeration<URL> mine = new ResourceEnumeration(name);
         Enumeration<URL> base;
-        if (parent != null && (!parentHasBeenSearched || parent != getParent())) {
+        if (parent != null && !skipParent) {
             // Delegate to the parent:
             base = parent.getResources(name);
-            // Note: could cause overlaps in case
-            // ClassLoader.this.parent has matches and
-            // parentHasBeenSearched is true
         } else {
-            // ClassLoader.this.parent is already delegated to for example from
-            // ClassLoader.getResources, no need:
             base = Collections.emptyEnumeration();
         }
         if (isParentFirst(name)) {


src/main/org/apache/tools/ant/AntClassLoader.java
```diff
@@ -936,9 +936,6 @@ private URL getUrl(String name) {
      * data (images, audio, text, etc) that can be accessed by class
      * code in a way that is independent of the location of the code.
      *
-     * <p>Would override getResources if that wasn't final in Java
-     * 1.4.</p>
```
basil, line 940: This comment was obsolete as of commit 17527b6490, which added this code:

```java
/** {@inheritDoc} */
@Override
public Enumeration<URL> getResources(String name) throws IOException {
    return getNamedResources(name);
}
```

src/main/org/apache/tools/ant/AntClassLoader.java
```diff
@@ -982,26 +979,21 @@ protected URL findResource(final String name) {
      *
      * @param name The resource name to search for.
      *             Must not be <code>null</code>.
-     * @param parentHasBeenSearched whether ClassLoader.this.parent
-     * has been searched - will be true if the method is (indirectly)
-     * called from ClassLoader.getResources
+     * @param searchParent whether to search the parent first - will be true if the method is
+     *     invoked from {@link #getResources(String)} or {@link #getNamedResources(String)} and
+     *     false if the method is invoked from {@link #findResources(String)}.
```
basil, line 984: This documentation is misleading as of commit 17527b6490.

Before commit 17527b6490, we didn't override `getResources(String)`. The base version from `ClassLoader` is roughly as follows:

```java
public Enumeration<URL> getResources(String name) throws IOException {
    if (parent != null) {
        // parent.getResources(name);
    } else {
        // getBootstrapResources(name);
    } else {
        // findResources(name);
    }
}
```

This was `final` in earlier versions of Java, so the `parentHasBeenSearched` semantics make sense. We didn't want to repeat this work if invoked from `ClassLoader#getResources(String)`. But as of commit 17527b6490 we now have:


```java
/** {@inheritDoc} */
@Override
public Enumeration<URL> getResources(String name) throws IOException {
    return getNamedResources(name);
}
```

So we're now no longer running `ClassLoader#getResources()`, only our own code. The only thing that still is necessary is to distinguish between our own `getResources()` (which is recursive and must check the parent first) and our own `findResources()` (which is non-recursive and must not check the parent first). A more straightforward way of representing this in the code is with a `searchParent` boolean that is the inverse of the older boolean.

basil:
To summarize, the root cause of this problem is the `|| parent != getParent()` logic and this associated comment:

> Note: could cause overlaps in case ClassLoader.this.parent has matches and parentHasBeenSearched is true.

The essence of this PR is just removing that code and comment. Renaming `parentHasBeenSearched` to `skipParent` is just to improve readability.

@jaikiran seems a bit nervous about this, so let me write up a proof of why this PR is safe. I dug up the original reason why the offending code was written. In [Ant Bug 30161](https://bz.apache.org/bugzilla/show_bug.cgi?id=30161), @jglick wrote the following:

> Note that there is a problem with the patch (which is I believe unavoidable unless you set `CL.parent` correctly, or override `CL.getResources` in JDK 1.5 where it seems to be nonfinal): if `CL.parent` (whatever that is) does have any matches for the resource name, and `ACL.parent` has matches for it that are in fact derived ultimately from `CL.parent`, `ACL.getResources` will produce duplicates. I put in a quick hack to avoid this in case `CL.parent` == `ACL.parent`. […]

This comment was written in 2004. The problem is that the "quick hack" doesn't work for Jenkins, which necessitated [Kohsuke's change](https://github.com/jenkinsci/jenkins/commit/9a2882dd70) ten years later in 2014. Now in 2021, Ant has clearly decided _not_ to set `ClassLoader.parent`, going the opposite way from Kohsuke (see [the latest comment in Ant Bug 35436](https://bz.apache.org/bugzilla/show_bug.cgi?id=35436)), but Ant has decided to override `ClassLoader.getResources` in commit 17527b6. The end result is the same: the "quick hack" is no longer needed.

There's nothing more permanent than a temporary solution.

I have convinced myself that this change is not only safe but desirable both logically and through empirical testing with Jenkins; feel free to ask me for clarification if you have any more questions.

jglick:
I barely recall 8fdf272627418e1e55adcfe8a52feb9125515be0 and do not have any special knowledge of the surrounding code here. Sounds OK offhand. Is it reasonable to add test coverage for this? Maybe with some special `ClassLoader` impl which counts resource lookups?

basil:
Probably, but I lack the time and interest to add a test. As I wrote above, I tested this in Jenkins and have convinced myself the fix is correct.

jaikiran:
Hello @basil, thank you for your detailed analysis and explanation of your changes. This looks good to me, plus since Jesse too has had a look at this gives me additional confidence on these changes. 

As for the tests, it would be good to add new test for this, but given the nature of the issue that's discussed in the linked Jenkins issue, I can understand that it may not be straightforward. We do have tests for `AntClassLoader` in our testsuite and I ran the `AntClassLoaderTest` against JDK 8, 11 and 17 EA on my local setup with these changes and it passed on those versions. So I think that's good enough to merge this PR.

Thank you for these changes.

PR #152, opened 2021-07-21
Call correct method to get parent from getRootLoader()

basil:
See [#149 (comment)](https://github.com/apache/ant/pull/149#issuecomment-884318912). Similar to what #151 did for `findResources(String, boolean)`, this PR amends `getRootLoader()` to invoke the correct method to get the parent. The `getParent()` method returns the immutable `parent` field from the parent `ClassLoader` class, which is not used by `AntClassLoader`. The correct method is `getConfiguredParent()` (introduced in commit 26f846b83), which returns the mutable `parent` field from the `AntClassLoader` class.

**Untested.**

### DIFF
--- a/src/main/org/apache/tools/ant/AntClassLoader.java
+++ b/src/main/org/apache/tools/ant/AntClassLoader.java
@@ -860,10 +860,18 @@ public class AntClassLoader extends ClassLoader implements SubBuildListener, Clo
      */
     private ClassLoader getRootLoader() {
         ClassLoader ret = getClass().getClassLoader();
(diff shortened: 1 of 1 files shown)


basil:
Closing this PR, as I don't think it makes sense anymore. The `getRootLoader()` method is only invoked when this class is running in "isolated" mode, which is documented as follows:

```java
/**
 * Whether or not this classloader will ignore the base
 * classloader if it can't find a class.
 *
 * @see #setIsolated(boolean)
 */
private boolean ignoreBase = false;
```

PR #157, opened 2021-08-31
Performance:

arturobernalg:
* Use arrayCopy
* Avoid redundant String.format
* Exit the loop early

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/compilers/JavacExternal.java
+++ b/src/main/org/apache/tools/ant/taskdefs/compilers/JavacExternal.java
@@ -94,8 +94,8 @@ public class JavacExternal extends DefaultCompilerAdapter {
         for(int i = nonArgFileOptionIdx + 1; i < commandLine.length; i++) {
             if (!isArgFileEligible(commandLine[i])) {
                 String option = commandLine[i];
-                for(int j = i - 1; j >= nonArgFileOptionIdx; j--) {
-                    commandLine[j + 1] = commandLine[j];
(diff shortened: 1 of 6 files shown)


src/main/org/apache/tools/ant/taskdefs/compilers/JavacExternal.java
```diff
@@ -94,9 +94,8 @@ private int moveArgFileEligibleOptionsToEnd(String[] commandLine) {
         for(int i = nonArgFileOptionIdx + 1; i < commandLine.length; i++) {
             if (!isArgFileEligible(commandLine[i])) {
                 String option = commandLine[i];
-                for(int j = i - 1; j >= nonArgFileOptionIdx; j--) {
-                    commandLine[j + 1] = commandLine[j];
-                }
+                if (i - nonArgFileOptionIdx >= 0)
```
bodewig, line 97: please add braces even for single line blocks

bodewig:
"performance" is such a big word, I very much doubt there will be any measurable effect on a real build by the changes :-)

anyway, thanks for the patch @arturobernalg . Please bring back the braces and please add yourself to the CONTRIBUTORS and contributors.xml files (unless you are already there and I just didn't see it).

src/main/org/apache/tools/ant/taskdefs/compilers/JavacExternal.java
arturobernalg: HI @bodewig 
I just add the braces.
TY

bodewig:
Thank you @arturobernalg

PR #161, opened 2021-10-02
Use Jakarta EE

twogee:
Java Mail became Jakarta Mail, and Ant should prepare for EOL of the former.


bodewig:
There might be a way to reduce the code duplication between the two mailer classes, but that can be done as a refactoring later. Also we'll need a new POM and a new artifact to publish. I'll look into this after 1.10.12 is out.

The change to ClasspathUtils looks good, thank you, but is not really related to this change.

twogee:
Thanks @bodewig, I missed the POM part. There's no need for a separate POM: I propose bumping the version in the current POM and adding a profile that overrides the dependencies in case somebody prefers the old javax.mail; I will add the change if the proposal is acceptable.

bodewig:
Personally I'd prefer two separate jars and two separate POMs.

twogee:
I believe this is a case of a clean change when only one of the two implementations is necessary (in contrast to the muddle of JAI vs ImageIO).

twogee:
I take back my previous proposal, I was thinking from a consumer POM perspective, which is incorrect. Marking MimeMailer as deprecated, hoping deprecated classes will be removed some day.

twogee:
Thanks for the review, could you please confirm if my understanding is correct that packaging some classes twice is acceptable?

bodewig:
I'm afraid we are talking past each other as I don't understand the question :-)

Current master creates ant-javamail.jar containing `MimeMailer` and its nested classes. I'd like to see the new `JakartaMailer` and its nested classes go into a separate jar named ant-jakartamail.jar (for example).

twogee:
@bodewig thanks again for helping with the confusion.

PR #162, opened 2021-10-02
Use Bulk operation instead of iteration.

arturobernalg:
Some operations inside loops that could be replaced with a bulk method. Bulk methods are shorter


src/main/org/apache/tools/ant/taskdefs/email/MimeMailer.java
```diff
@@ -294,7 +294,7 @@ public void send() {
                 ? new InternetAddress(addr.getAddress())
                 : new InternetAddress(addr.getAddress(), name));
         }
-        return addrs.toArray(new InternetAddress[addrs.size()]);
+        return addrs.toArray(new InternetAddress[0]);
```
bodewig, line 297: seems to belong to a separate PR

bodewig (changes requested):

src/main/org/apache/tools/ant/taskdefs/email/MimeMailer.java
arturobernalg: Hi @bodewig 
Reverted.
TY

bodewig:
thank you @arturobernalg Currently there is a release vote underway and I don't want to disturb things by changing `master` before it is done.

arturobernalg:
> thank you @arturobernalg Currently there is a release vote underway and I don't want to disturb things by changing `master` before it is done.

Completely agree. This is a minor change. can wait. 
Thank you @bodewig

PR #163, opened 2021-10-02
Make final var  when is possible.

arturobernalg:


src/main/org/apache/tools/ant/taskdefs/AbstractJarSignerTask.java
```diff
@@ -381,7 +381,7 @@ protected void setCommonOptions(final ExecTask cmd) {
      * @throws BuildException if the property is not correctly defined.
      */
     protected void declareSysProperty(
-        ExecTask cmd, Environment.Variable property) throws BuildException {
+            final ExecTask cmd, final Environment.Variable property) throws BuildException {
```
bodewig, line 384: why did you change indentation here?

bodewig:
This is a huge change when it comes to the number of lines touched. When faced with such a big change I always fear I may overlook something subtle.

Personally I don't believe making the argument of a one-line method that just calls another method passing on the argument is improving anything. YMMV.

arturobernalg:
> This is a huge change when it comes to the number of lines touched. When faced with such a big change I always fear I may overlook something subtle.
> 
> Personally I don't believe making the argument of a one-line method that just calls another method passing on the argument is improving anything. YMMV.

HI @bodewig 
I close the PR
TY

bodewig:
@arturobernalg I'm by far not the only Ant committer, so please don't feel forced to close a PR just because I comment in an unfavorable way.

arturobernalg:
> @arturobernalg I'm by far not the only Ant committer, so please don't feel forced to close a PR just because I comment in an unfavorable way.

HI @bodewig 
TY for the feedback. I'll reopen the PR.
TY

PR #164, opened 2021-10-02
Use inner class as static.

arturobernalg:
Its not require access to an enclosing instance's non-public fields and methods.

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/Input.java
+++ b/src/main/org/apache/tools/ant/taskdefs/Input.java
@@ -44,7 +44,7 @@ public class Input extends Task {
     /**
      * Represents an InputHandler.
      */
-    public class Handler extends DefBase {
+    public static class Handler extends DefBase {
 
         private String refid = null;
         private HandlerType type = null;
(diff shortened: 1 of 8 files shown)


src/main/org/apache/tools/ant/taskdefs/Input.java
```diff
@@ -44,7 +44,7 @@
     /**
      * Represents an InputHandler.
      */
-    public class Handler extends DefBase {
+    public static class Handler extends DefBase {
```
bodewig, line 47: this change is not backwards compatible

src/main/org/apache/tools/ant/taskdefs/Javac.java
```diff
@@ -910,7 +910,7 @@ public boolean getNowarn() {
      */
     public ImplementationSpecificArgument createCompilerArg() {
         final ImplementationSpecificArgument arg =
-            new ImplementationSpecificArgument();
+                new ImplementationSpecificArgument();
```
bodewig, line 913: please don't change whitespace in existing code.

bodewig:
Thank you @arturobernalg but I'm afraid this is a change that is not backwards compatible at all so we cannot accept it.

arturobernalg:
> Thank you @arturobernalg but I'm afraid this is a change that is not backwards compatible at all so we cannot accept it.

HI @bodewig 
Ok, close the PR. maybe for a new. mayor release.
TY

bodewig:
for historical reasons Ant is not following semantic versioning, so the next "major release" would be 1.11.0. Having said that I'm not aware of anybody planning to create such a thing.

PR #166, opened 2021-10-30
Support Jakarta Mail in MailLogger

twogee:
Unfortunately, the code is duplicated due to different logging in different objects.

### DIFF
--- a/src/main/org/apache/tools/ant/listener/MailLogger.java
+++ b/src/main/org/apache/tools/ant/listener/MailLogger.java
@@ -47,12 +47,12 @@ import org.apache.tools.mail.MailMessage;
  *  <ul>
  *    <li> MailLogger.mailhost [default: localhost] - Mail server to use</li>
  *    <li> MailLogger.port [default: 25] - Default port for SMTP </li>
- *    <li> Maillogger.user [no default] - user name for SMTP auth
- *    (requires JavaMail)</li>
- *    <li> Maillogger.password [no default] - password for SMTP auth
- *    (requires JavaMail)</li>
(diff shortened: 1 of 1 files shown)


bodewig:
In the special case of a logger it may be even preferable to not log any message at all as the message would in turn be logged to the same logger again and may never become visible anywhere. I see the existing code has a `log` statement but I wonder whether this is just an oversight. There are usages of `System.err/out` which may be the better choice.

twogee:
I reason that it might be preferable to have those messages in the log at higher verbosity. Would adding a new interface to provide implementations of log create backwards compatibility problems?

bodewig:
given this all happens in `buildFinished` and it will not be called again after that, my fear of causing a cycle is unfounded.

Wade-tech:
Hi ,tomcat10 change javax.* namespace to jakarta.*， Does ant now support jakarta.mail?

bodewig:
https://ant.apache.org/manual/Tasks/mail.html - Ant 1.10.13.

Wade-tech:
Thanks, JakartaMimeMailer.java use jakarta.mail . I downdload org.apache.ant-ant-1.10.13 jar(url:https://mvnrepository.com/artifact/org.apache.ant/ant/1.10.13). In this jar , not found JakartaMimeMailer.java , i'm confused
image

jaikiran:
Hello @Wade-tech, it will be present in the ant-jakartamail artifact which is here  https://repo1.maven.org/maven2/org/apache/ant/ant-jakartamail/1.10.13/

Wade-tech:
Thanks

PR #167, opened 2021-11-01
Clarify logging

twogee:


twogee:
Hopefully these changes make logging methods a bit more comprehensible.
They are backwards binary compatible (according to japicmp 0.15.4).

bodewig:
I may have become used to the existing logging setup, but to be honest your proposed change doesn't make anything more comprehensible to me. That may just be me.

twogee:
The changes make it clear which objects control logging.

These are DefaultLogger (which currently has a protected log method), Project and ProjectComponent (a superclass of Task, so the latter simply inherits its log method). TraxLiaison sets a Logger, no need for the XSLTLogger.

The next step will be to remove the code duplication discussed in #166 by moving the getMailerImplementation to Mailer and passing a Logger to it.

PR #169, opened 2021-11-15
junitlauncher - Improved extensibility capabilities for formatters

azotcsit:
### Context
I'm working on migration of Cassandra to JUnit5 (https://issues.apache.org/jira/browse/CASSANDRA-16630). Our test running code is heavily customized and based on custom formatters ([XML](https://github.com/apache/cassandra/blob/trunk/test/unit/org/apache/cassandra/CassandraXMLJUnitResultFormatter.java), [Brief](https://github.com/apache/cassandra/blob/trunk/test/unit/org/apache/cassandra/CassandraBriefJUnitResultFormatter.java)). In fact these are just extensions of the default _ant-junit_ formatters. _ant-junitlauncher_ provides similar default implementations called _legacy-xml_ and _legacy-brief_. Unfortunately, we cannot extend and modify these formatters to incorporate our custom logic. The purpose of this PR is to re-use logic from `AbstractJUnitResultFormatter` class in custom existing implementations.

### Behavior
This PR does not change any existing behavior.

### Summary of the changes
1. Made `AbstractJUnitResultFormatter` class public to let it be extended
2. Made methods protected, so they can be customized in children classes

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
@@ -43,7 +43,7 @@ import java.util.Optional;
 /**
  * Contains some common behaviour that's used by our internal {@link TestResultFormatter}s
  */
-abstract class AbstractJUnitResultFormatter implements TestResultFormatter {
+public abstract class AbstractJUnitResultFormatter implements TestResultFormatter {
 
     protected TestExecutionContext context;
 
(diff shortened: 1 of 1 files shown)


src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/AbstractJUnitResultFormatter.java
```diff
@@ -43,7 +43,7 @@
 /**
  * Contains some common behaviour that's used by our internal {@link TestResultFormatter}s
  */
-abstract class AbstractJUnitResultFormatter implements TestResultFormatter {
+public abstract class AbstractJUnitResultFormatter implements TestResultFormatter {
```
azotcsit, line 46: It is impossible to have abstract class protected, so it is marked as public. The methods are protected, not public.

azotcsit:
Closing as per the discussion on the dev mail list: https://lists.apache.org/thread/2v8l95fwh0sgz0vlgqcf1rpd88swf3pv

PR #173, opened 2021-12-20
Allow ant:get task to disable authentication on redirect.

bernolanger:
Most clients do not send the Authorization header on redirects by default; because of security issues.

The ant:get task instead, always sends the Authorization header to the redirected location.

This PR makes this behavior configurable. The optional attribute "authenticateOnRedirect" can be set to "false".

I'm not a security expert. Therefore I didn't change the default behavior to avoid breaking existing Ant scripts. This means, "authenticateOnRedirect" defaults to "true". But maybe it would be better to change this.

_Example: getting an artifact from AWS CodeArtifact which redirects to a pre signed URL and therefore mustn't contain the Authorization header:_
```
<get src="https://codeartifact-url/..." username="aws" password="<secret>" dest="..." authenticateOnRedirect="false">
  <header name="Accept" value="*/*"/>
</get>
```

### DIFF
--- a/manual/Tasks/get.html
+++ b/manual/Tasks/get.html
@@ -92,6 +92,12 @@ the request is relayed to the proxy.</p>
     <td>Yes if <var>username</var> is set</td>
   </tr>
   <tr>
+    <td>authenticateOnRedirect</td>
+    <td>Whether the credentials should also be sent to the new location when a redirect is followed.<br/>
+    <em>since Ant 1.10.13</em></td>
+    <td>No; default is <q>true</q></td>
+  </tr>
+  <tr>
     <td>maxtime</td>
     <td>Maximum time in seconds a single download may take, otherwise it will be interrupted and
       treated like a download error.  <em>Since Ant 1.8.0</em></td>
--- a/src/main/org/apache/tools/ant/taskdefs/Get.java
+++ b/src/main/org/apache/tools/ant/taskdefs/Get.java
@@ -82,6 +82,7 @@ public class Get extends Task {
     private boolean ignoreErrors = false;
     private String uname = null;
     private String pword = null;
+    private boolean authenticateOnRedirect = true; // on by default for backward compatibility
     private long maxTime = 0;
     private int numberRetries = NUMBER_RETRIES;
     private boolean skipExisting = false;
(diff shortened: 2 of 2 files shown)


bodewig:
many thanks @bernolanger

we'd like to credit you in CONTRIBUTORS and contributors.xml. What is the name you'd want us to use?

jaikiran:
>Therefore I didn't change the default behavior to avoid breaking existing Ant scripts. This means, "authenticateOnRedirect" defaults to "true". But maybe it would be better to change this.

I was leaning towards making this new `authenticateOnRedirect` to default to `false` to be more secure (i.e. don't set Authorization header to redirected URL unless explicitly asked to). That might break scripts but I think that's probably a good thing since it would force users to review their target URLs and decide if they really want to send the auth header on redirect for that specific URL.

The only place where this would probably be a nuisance is if the redirect is happening just for the scheme. What I mean is if the original URL `http://example.com/foo` was redirecting to `https://example.com/foo`. Or even in some cases where servers redirect a URL of the form `http://example.com/foo` to `http://example.com/foo/` (slash at the end). So yes, I guess leaving the current backward compatible behaviour is OK.

bodewig:
you are  right @jaikiran, I'll change the default.

sfllaw:
82f5edcf49002943773a165464e152b1d75a68a8 changed the default value of `authenticateOnRedirect` from true to false, which is backwards incompatible but is actually proper behaviour.

sfllaw:
I encountered this bug when using `<get>` with HTTP Basic Auth to supply a username and password. If the server responds with a 3xx redirect to an S3 bucket, Amazon will respond with a 400 Bad Request:
> Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified

I’m leaving this comment so that others can find this issue. Ant 1.10.13 will probably solve this, once it is released.

See also:
* [Bad Request (400) from S3 when downloading artifact from TeamCity in "ant get" task using basic http auth scheme (httpAuth prefix in URL)
](https://github.com/JetBrains/teamcity-s3-artifact-storage-plugin/tree/0dcc5c04d514cc23e7369b7e092924f67abfc7e8#bad-request-400-from-s3-when-downloading-artifact-from-teamcity-in-ant-get-task-using-basic-http-auth-scheme-httpauth-prefix-in-url)

PR #174, opened 2021-12-22
junitlauncher - Added fork mode support

azotcsit:
### Context
I'm working on migration of Cassandra to JUnit5 (https://issues.apache.org/jira/browse/CASSANDRA-16630). Our test running code is heavily customized and relies onto [forkMode](https://github.com/apache/ant/blob/master/src/main/org/apache/tools/ant/taskdefs/optional/junit/JUnitTask.java#L185) property. Unfortunately, _junitlauncher_ task does not have any equivalent. The purpose of this change is to introduce a way to run tests independently (in separate JVMs).

### Behavior
This PR does not change any existing behavior. By default, the tests will be run in the same forked JVM (aka mode="once").

### Summary of the changes
1. Added `node` attribute to `fork` element
2. Added logic to spin a JVM per test suite class into `JUnitLauncherTask`
3. Updated tests and documentation

**I'm not sure `JUnitLauncherTaskTest` is run during `./build clean test`, hence the test might be broken.**


slachiewicz:
This looks superseded — the capability landed in 1.10.14.

Commit f7f3bdfa033b2c77e001c2fc485bf200a682227d (Bugzilla 65176) added a `forkMode`
attribute on `<fork>` with the value `perTestClass`, which launches each test class in a
separate JVM. That is the behaviour this PR proposed as `mode="perTest"`.

The spelling differs, so anything written against this PR needs adjusting:

| here | shipped |
| --- | --- |
| `mode="perTest"` | `forkMode="perTestClass"` |
| `mode="once"` (explicit default) | omit `forkMode` |

`ForkDefinition$ForkMode` is absent from `ant-junitlauncher` 1.10.13 and present in
1.10.14 through 1.10.17.

Worth closing, unless you wanted `once` to remain an explicit value.

*This comment was created with AI assistance.*

PR #184, opened 2022-04-05
Bugs are Fixed

sm3533:


bodewig:
I'm not really sure what to make of this:

* please don't change the format of files. Either change whitespace or code but not both. Otherwise the PR creates too much noise to tell the real changes from cosmetics. Having said that, why would anybody want a blank line between the javadoc block and the method declaration?
* the change to `PropertyHelper` is just wrong, it intentionally returns `null` - see javadoc
* I don't understand the only non-cosmetic change to `Locator`. Why is the current condition of `>= 0` wrong?
* the `synchronized` block in `Parallel` is there for a reason, see the comment
* `DeweyDecimal` is a public class, removing it breaks backwards compatibility
* the `final` in `Sound` may or may not be a good idea, but why do you remove the comment?
* I don't think the change to `LayoutPreservingProperties` improves readability

Overall there are too many unrelated changes in a single PR to make this useful, even if I agreed with all of them. Please create pull requests that do one thing and one thing only.

PR #191, opened 2022-08-04
Fix retrieval from non-default repositories

ppkarwasz:
Since the `<remoterepo>` tag is not in the scope of the `<resolve>` target, fetching artifacts from repositories different from Maven
Central fails (e.g. JAI).

### DIFF
--- a/fetch.xml
+++ b/fetch.xml
@@ -119,8 +119,8 @@ Set -Ddest=LOCATION on the command line
             </not>
           </condition>
         </fail>
-        <resolver:remoterepo url="@{repository}" id="@{id}"/>
         <resolver:resolve>
+          <resolver:remoterepo url="@{repository}" id="@{id}"/>
           <dependencies id="@{archive}.path">
             <dependency groupId="@{project}"
                         artifactId="@{archive}"


jaikiran:
Hello @ppkarwasz, thank you for this change. It looks OK to me, but I'm surprised that without this change, the fetch fails for you. I tested it locally (without this change) and our CI instance have a job which runs the fetch command and that seems to be passing. 
Do you have a command and log output that has more details about the failure?

bodewig:
I can reproduce this locally and believe the key is to remove jai from the local m2 repo. Maybe the CI server has stored jai there as well @jaikiran ?

bodewig:
@ppkarwasz many thanks, what name do you want us to use when we add you to the contributors file?

ppkarwasz:
@bodewig, I didn't notice I never filled up the Github profile with my real name. You can use `Piotr P. Karwasz`.

bodewig:
see b9272756f

Even if your name had been present I would have asked :-) . Not all people want their name to show up in a list like ours.

jaikiran:
> I can reproduce this locally and believe the key is to remove jai from the local m2 repo. Maybe the CI server has stored jai there as well @jaikiran ?

That's very likely. Thank you @ppkarwasz for the fix.

PR #194, opened 2022-10-26
Names end before the first NULL (not the last)

keithc-ca:
This fixes parsing of archives produced on macOS.
See the discussion in https://github.com/ibmruntimes/Semeru-Runtimes/issues/15.

### DIFF
--- a/src/main/org/apache/tools/tar/TarUtils.java
+++ b/src/main/org/apache/tools/tar/TarUtils.java
@@ -286,9 +286,9 @@ public class TarUtils {
                                    final ZipEncoding encoding)
         throws IOException {
 
-        int len = length;
-        for (; len > 0; len--) {
-            if (buffer[offset + len - 1] != 0) {
+        int len = 0;
+        for (; len < length; ++len) {
+            if (buffer[offset + len] == 0) {
                 break;
             }
         }


qf28:
Hi

bodewig:
I should have asked my five year younger self wether I remember why the code is what it is :-) - see https://github.com/apache/commons-compress/pull/54

I'll merge this PR right away. We'd like to credit you in CONTRIBUTORS and contributors.xml if you want to. Which name would you want us to add?

bodewig:
thank you!

bodewig:
actually this breaks support for file names using multi-byte encodings where NULs may just be part of a multi-byte sequence and not signal the end of the name. You can see this by running 

```
/build.sh -f src/etc/testcases/taskdefs/untar.xml encodingTest
```

in Ant's root. This creates a tar using UTF-16 and fails to expand it again. I'm not 100% sure how to solve this as `ZipEncoding` might be a bit too simplistic.

keithc-ca:
> `ZipEncoding` might be a bit too simplistic

Yes, I think so. I explored the idea of just decoding the whole byte array, but that will fail if there's a bad number of nulls at the end (e.g. an odd number for UTF16).

bodewig:
well, in the end the tar format is not meant to store file names in the name field with a multi-byte encoding. The few docs that one can find talk about "local variant of ASCII" and the UTF16 test only worked by accident before the change.

I'll change the docs in either case, discouraging the use of multi-byte encodings and change the test to use codepage 1252 or something like that. And then I'll try to think about a proper - ideally backwards compatible - extension of ZipEncoding that would  help us here.

bodewig:
it seems e04fbe7ff works for our tests. I thought about letting the encoding tell us what a NUL would look like and then search for that, this seems to work even though it was more complex than I hoped for as Java's UTF-16 encoder prepends a BOM to every string, including one only consisting of a NUL.

Of course this is slower than the code we had before but parsing tar archive names is not likely to be the performance bottleneck of a build process. And people should not use Ant's tar package but rather switch to Commons Compress outside of the context of Ant itself.

bodewig:
Good catch, thank you.

I could try to turn the byte array into a string, catch the exception and then keep searching for encoded NULs later.

But even then it might just fail for some other encoding, something that's even more complex and uses different numbers of bytes per glyph like UTF-8. The more I think about it the more I feel it is impossible to make the code work for general multi-byte encodings, at least without moving the burden to `ZipEncoding` itself. Will need to think about that a bit more.

PR #195, opened 2022-11-16
Fix detection of java.specification.version

keithc-ca:
The existing use of `tr` doesn't distinguish between Java 8 and Java 18; see https://github.com/apache/ant/commit/bbe6859b2f57a2c9dcb51e269527f3c18a74aa1d#commitcomment-90131350 - this fixes that.

### DIFF
--- a/src/script/ant
+++ b/src/script/ant
@@ -369,11 +369,18 @@ else
   fi
 fi
 # Run "java -XshowSettings:properties" and check the output for "java.specification.version" value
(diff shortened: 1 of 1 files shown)


jaikiran:
Thank you @keithc-ca for this change. I will run some CI test before merging (I'll also check why my testing of my previous commit didn't catch the 1.8 version issue).

jaikiran:
> I'll also check why my testing of my previous commit didn't catch the 1.8 version issue

For the record - one of the test jobs did catch this issue for 1.8 version, I just hadn't caught up with my email notifications for these failures.

Thank you again for this fix.

PR #197, opened 2023-01-04
Handling forked VM timeout in a similar way to junit ant task

tkowalcz:
so that settings like haltonfailure have effect. This strives to make junitlauncher task behave similar to junit (4) task.

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/confined/JUnitLauncherTask.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/confined/JUnitLauncherTask.java
@@ -330,7 +330,20 @@ public class JUnitLauncherTask extends Task {
                 break;
             }
             case Constants.FORK_EXIT_CODE_TIMED_OUT: {
-                throw new BuildException(new TimeoutException("Forked test(s) timed out"));
+                // test has failure(s)
+                try {
+                    if (test.getFailureProperty() != null) {
+                        // if there are test failures and the test is configured to set a property in case
+                        // of failure, then set the property to true
+                        this.getProject().setNewProperty(test.getFailureProperty(), "true");
+                    }
+                } finally {
+                    if (test.isHaltOnFailure()) {
+                        throw new BuildException(new TimeoutException("Forked test(s) timed out"));
+                    } else {
+                        log("Timeout occurred. Please note the time in the report does not reflect the time until the timeout.");
+                    }
+                }
             }
         }
     }


src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/confined/JUnitLauncherTask.java
```diff
@@ -330,7 +330,20 @@ private void forkTest(final TestDefinition test) {
                 break;
             }
             case Constants.FORK_EXIT_CODE_TIMED_OUT: {
-                throw new BuildException(new TimeoutException("Forked test(s) timed out"));
+                // test has failure(s)
+                try {
+                    if (test.getFailureProperty() != null) {
+                        // if there are test failures and the test is configured to set a property in case
```
jaikiran, line 336: Minor nit - I think this comment should say "if a test timed out ...."

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/confined/JUnitLauncherTask.java
```diff
@@ -330,7 +330,20 @@ private void forkTest(final TestDefinition test) {
...
+                        // of failure, then set the property to true
+                        this.getProject().setNewProperty(test.getFailureProperty(), "true");
+                    }
+                } finally {
+                    if (test.isHaltOnFailure()) {
+                        throw new BuildException(new TimeoutException("Forked test(s) timed out"));
+                    } else {
+                        log("Timeout occurred. Please note the time in the report does not reflect the time until the timeout.");
```
jaikiran, line 344: Perhaps remove the report part from this log message and instead just say `log("Some test(s) timed out")`?

jaikiran:
Thank you Tomasz for this PR. I think this change looks fine. Could you add a test case to verify this change works as expected. We have a junitlauncher test class at `src/tests/junit/org/apache/tools/ant/taskdefs/optional/junitlauncher/JUnitLauncherTaskTest.java` which uses the `src/etc/testcases/taskdefs/optional/junitlauncher.xml` as the build file to test. Perhaps you could add a new test method to it to try a test which times out intentionally after a few seconds?

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/confined/JUnitLauncherTask.java
tkowalcz: I copied the message as it appears in junit4. I don't have opinion one way or the other.

tkowalcz:
Thanks for review. I will make changes as you suggested and of course write a test.

jaikiran:
Hello Tomasz, I don't want to rush you, but if writing a test is proving to be time consuming or complicated, do let me know and I'll merge this fix (which looks good to me) and I'll add a test myself.

tkowalcz:
Hi, actually I had problems running the tests in the first place. When I build the project the tests do not seem to be executed (build pass even when I introduce failing assertion). I can later today copy exactly which steps I made.

jaikiran:
> When I build the project the tests do not seem to be executed 

Some of the tasks in Ant, including the `junitlauncher` aren't part of the "core". You will have to first fetch the dependencies of these tasks as follows:
```
ant -f fetch.xml -Ddest=optional
```
and then run:
```
./build.sh clean test
```

tkowalcz:
Thanks. I will give it a try and let you know.

tkowalcz:
@jaikiran I'm sorry but I was unable to run the tests. After implementing your instructions the `LegacyXmlResultFormatterTest` was run but there was no mention of `JUnitLauncherTaskTest` being run. I tried investigating this on multiple fronts but still have no idea how to run it as part of the suite.

jaikiran:
The change in this PR looks fine to me. A test can be added separately - I haven't found the time to try and help you get the test implemented/run. I'll go ahead and merge this now and will add the test later.

PR #198, opened 2023-02-02
Using static modifier to normalize methods contract and performance

tbw777:


bodewig:
I must say I find it extremely difficult to properly review a PR that touches 100+ files, even if the change is supposed to be mechanical.

In this case I saw a change to `Antlib.java` where a string literal was replaced by a constant defined somewhere else in the class - a place that is not visible inside the diff so I need to look up its definition elsewhere. This not only makes the review more difficult it also means I can not just scroll down the page and press merge as the set of changes obviously also contains changes not described by the headline.

All in all I don't see myself being able to review this in any reasonable time frame. If you want to see this applied, please cut it into smaller chunks.

PR #200, opened 2023-04-11
Do not treat failure of `System.setSecurityManager` as fatal

jglick:
Fixes https://github.com/spotbugs/spotbugs-maven-plugin/issues/423. When running on Java 18+, where `System.setSecurityManager` is not allowed by default, just let in-process Java execution run without a security manager. Normally this is not a big deal; it would only matter if you were running a program which actually tried to `System.exit`, which is much better handled by simply forking.

Verified by applying this patch as well as

```diff
diff --git src/etc/poms/ant/pom.xml src/etc/poms/ant/pom.xml
index 9a96764be..b062bbe63 100644
--- src/etc/poms/ant/pom.xml
+++ src/etc/poms/ant/pom.xml
@@ -343,7 +343,7 @@
     <testResources>
       <testResource>
         <directory>../../../../src/etc/testcases</directory>
-        <filtering>true</filtering>
+        <filtering>false</filtering>
       </testResource>
       <testResource>
         <directory>../../../../src/main</directory>
```

(without which I get errors processing various binary or non-UTF-8 test resources; is this code path not tested during release?) and

```bash
mvnd -f src/etc/poms -am -pl ant -DskipTests install
```

and then patching `spotbugs/spotbugs-maven-plugin` with

```diff
diff --git pom.xml pom.xml
index 3a95e22..ca4ef56 100644
--- pom.xml
+++ pom.xml
@@ -160,7 +160,7 @@
     <spotbugsTag>4.7.3</spotbugsTag>
     <slf4jVersion>2.0.7</slf4jVersion>
 
-    <antVersion>1.10.13</antVersion>
+    <antVersion>1.10.14-SNAPSHOT</antVersion>
     <groovyVersion>4.0.11</groovyVersion>
     <javaparserVersion>3.25.2</javaparserVersion>
 
```

and

```bash
mvn -DskipTests install
```

and then in `jenkinsci/scm-api-plugin` running

```bash
JAVA_HOME=/usr/lib/jvm/java-19-openjdk-amd64 mvn spotbugs:check -Dspotbugs.fork=false -Dspotbugs-maven-plugin.version=4.7.3.5-SNAPSHOT -e
```

(I initially tried to write an IT in `spotbugs-maven-plugin` but failed to reproduce the error there, for reasons unknown.)

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/ExecuteJava.java
+++ b/src/main/org/apache/tools/ant/taskdefs/ExecuteJava.java
@@ -151,6 +151,7 @@ public class ExecuteJava implements Runnable, TimeoutObserver {
                 throw new BuildException(
                     "main() method in %s is not declared static", classname);
             }
+            perm.setProject(project);
             if (timeout == null) {
                 run(); //NOSONAR
             } else {
@@ -201,6 +202,7 @@ public class ExecuteJava implements Runnable, TimeoutObserver {
             if (sysProperties != null) {
                 sysProperties.restoreSystem();
             }
+            perm.setProject(null);
         }
     }
 
--- a/src/main/org/apache/tools/ant/types/Permissions.java
+++ b/src/main/org/apache/tools/ant/types/Permissions.java
@@ -30,6 +30,7 @@ import java.util.StringTokenizer;
 
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.ExitException;
+import org.apache.tools.ant.Project;
 
 /**
  * This class implements a security manager meant for usage by tasks that run inside the
@@ -52,6 +53,7 @@ public class Permissions {
     private SecurityManager origSm = null;
     private boolean active = false;
     private final boolean delegateToOldSM;
+    private Project project = null;
 
     // Mandatory constructor for permission object.
     private static final Class<?>[] PARAMS = {String.class, String.class};
@@ -72,6 +74,10 @@ public class Permissions {
      */
     public Permissions(final boolean delegateToOldSM) {
         this.delegateToOldSM = delegateToOldSM;
+    }
+
+    public void setProject(Project project) {
+        this.project = project;
     }
 
     /**
@@ -100,8 +106,14 @@ public class Permissions {
     public synchronized void setSecurityManager() throws BuildException {
         origSm = System.getSecurityManager();
         init();
-        System.setSecurityManager(new MySM());
-        active = true;
+        try {
+            System.setSecurityManager(new MySM());
+            active = true;
+        } catch (UnsupportedOperationException x) {
+            if (project != null) {
+                project.log("Skipping security manager (perhaps running on Java 18+)", x, Project.MSG_INFO);
+            }
+        }
     }
 
     /**
@@ -169,8 +181,10 @@ public class Permissions {
      * To be used by tasks that just finished executing the parts subject to these permissions.
      */
     public synchronized void restoreSecurityManager() {
-        active = false;
-        System.setSecurityManager(origSm);
+        if (active) {
+            active = false;
+            System.setSecurityManager(origSm);
+        }
     }
 
     /**


jglick:
> the work we had done in a recent release to set `java.security.manager=allow` when launching Ant

82c70f3 I guess. Yes I think it would be appropriate to not even attempt to call `setSecurityManager` in recent Java releases; the original use case of running a program in-JVM which calls `System.exit` is rather dubious.

mabartos:
@jaikiran Hello, any progress with this issue? Or is there any workaround for this?

stevenschlansker:
I also ran into this issue while running SpotBugs with `<fork>false</fork>`:

```
Caused by: org.apache.tools.ant.BuildException: java.lang.UnsupportedOperationException: The Security Manager is deprecated and will be removed in a future release
    at org.apache.tools.ant.taskdefs.ExecuteJava.execute (ExecuteJava.java:194)
    at org.apache.tools.ant.taskdefs.Java.run (Java.java:891)
    at org.apache.tools.ant.taskdefs.Java.executeJava (Java.java:231)
    at org.apache.tools.ant.taskdefs.Java.executeJava (Java.java:135)
    at org.apache.tools.ant.taskdefs.Java.execute (Java.java:108)
    at org.apache.tools.ant.UnknownElement.execute (UnknownElement.java:299)
    at jdk.internal.reflect.DirectMethodHandleAccessor.invoke (DirectMethodHandleAccessor.java:104)
    at java.lang.reflect.Method.invoke (Method.java:578)
    at org.apache.tools.ant.dispatch.DispatchUtils.execute (DispatchUtils.java:99)
    at groovy.ant.AntBuilder.performTask (AntBuilder.java:347)
    at groovy.ant.AntBuilder.nodeCompleted (AntBuilder.java:286)
    at groovy.util.BuilderSupport.doInvokeMethod (BuilderSupport.java:161)
    at groovy.ant.AntBuilder.doInvokeMethod (AntBuilder.java:219)
    at groovy.util.BuilderSupport.invokeMethod (BuilderSupport.java:75)
```

The security manager is going away. While it would be nice to not even try to set the security managers on newer JVMs, I think this patch would still be a step in the right direction.

jaikiran:
Sorry everyone, my testing with the changes I have is taking longer than expected. @stevenschlansker, are you running these tests by building Ant from source? If yes, then I'll go ahead and merge this PR since that will help your case for the time being.

jglick:
> are you running these tests by building Ant from source?

@jaikiran why would you ask that? https://github.com/apache/ant/pull/200#issuecomment-1631658175 reports an error that looks identical to that long reported in https://github.com/spotbugs/spotbugs-maven-plugin/issues/423 [using an Ant release](https://github.com/spotbugs/spotbugs-maven-plugin/blob/e8d61beed8f78772d3ed764f1542f98fffec1bc9/pom.xml#L177). Perhaps I am missing something.

stevenschlansker:
No, I am running the latest spotbugs-maven-plugin release, which uses Ant as a library. The problem only manifests if you disable forking in the plugin configuration.

jaikiran:
Hello Jesse,
> > are you running these tests by building Ant from source?
> 
> @jaikiran why would you ask that? [#200 (comment)](https://github.com/apache/ant/pull/200#issuecomment-1631658175) 

The reason I asked that question was to see if I should just go ahead and push this PR to master so that Steven could benefit from using the Ant version he builds from source to temporarily get past the issue he's running into. Instead of waiting for my bigger change. If Steven isn't using Ant built from source, then merging this wouldn't be of use to them since it would anyway require an Ant release.

jglick:
> it would anyway require an Ant release

Yes, and then for `spotbugs-maven-plugin` to update to that release and cut its own release.

jaikiran:
Hello Jesse and Steven, I've pushed some commits to master branch which reverts all the complex/brittle workarounds we had added in Ant 1.10.13 to keep using SecurityManager.

A fresh set of commits have now been made to not use SecurityManager in Ant in Java 18+ versions (Java 18 is the first release which by defaults disallows setting the SecurityManager).

If anyone of you builds Ant from source using recent Java versions, please go ahead and build it and try it out against your projects. I'll run some more final tests tomorrow and send out a mail to Ant dev mailing list about these changes and requesting everyone to start testing their projects with these changes.
 
I think you can now close this PR.

jaikiran:
Ant 1.10.14 (containing this change) is in voting phase if you would like to try it https://lists.apache.org/thread/th6yrxc4dtpqhc2vnzmrd7d9b881hk4g

stevenschlansker:
Thank you very much :)

PR #203, opened 2023-08-30
bz-67082 fix unintended trimmed string output by pathconvert

sratz:
PropertyOutputStream by default trims output.

This is undesired in the <pathconvert> use case. Explicitly set it to false.

Fixes regression introduced in 1.10.13 / 90ed3ff6cca8634e38d7c3c82858ce48c9c4be2b.


jaikiran:
Hello Sebastian, thank you for reporting this issue and providing a fix that too with a test case. This is a very well done patch. I've merged this and also included your name in our contributors list. The merged commit is https://github.com/apache/ant/commit/f3796ab1af7c6481553b758962cf7286099aaa67

PR #204, opened 2023-09-11
bz-66552 fix Depend task does not handle Dynamic constant pool entries - java.lang.ClassFormatError: Invalid Constant Pool entry Type 17

jmichelberger:
In cases the Depend task scans a class file with Constant Pool entry Type 17 there was no handling, just fireing an Exception.
 Constant is defined since JDK11, and format is same as InvokeDynamic Type 18.
 Provide Identical handling.


jaikiran:
Thank you @jmichelberger for this patch, this looks OK to me and I've now merged it and I have added you to our contributors list. The patch doesn't have  a test case, I'm guessing that you were however able to test these changes manually against the original project that you ran into an issue with. Is that right?

PR #206, opened 2023-11-15
Throw BuildExceptions in ReplaceRegExp for IOExceptions and missing files

mfaquan:
Currently, if an IOException occurs in ReplaceRegExp or a file is missing, the task logs the error, but does not fail.  This pull request has ReplaceRegExp throw a BuildException if the task encounters an IOException or a file is missing.

A failOnError attribute is also added that only affects these newly thrown BuildExceptions.  Existing BuildExceptions that were thrown won't be affected by the failOnError attribute, and backwards compatibility can be maintained with failOnError=false.

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/optional/ReplaceRegExp.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/ReplaceRegExp.java
@@ -481,10 +481,13 @@ public class ReplaceRegExp extends Task {
                 log("An error occurred processing file: '"
                     + file.getAbsolutePath() + "': " + e.toString(),
                     Project.MSG_ERR);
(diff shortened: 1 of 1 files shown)


bodewig:
You are correct and I don't recall why the code works the way it does. But it has done so for more than 20 years now and changing it might break existing builds. Let me suggest to introduce a `failOnError` attribute like some other tasks have - and make it default to `false` for the sake of backwards compatibility.

bodewig:
looks good, thank you. We'll also need to change the manual page, which I can do myself.

If you want to be listed in Ant's contributor's list, please let us know which name you want us to use.

mfaquan:
> looks good, thank you. We'll also need to change the manual page, which I can do myself.
> 
> If you want to be listed in Ant's contributor's list, please let us know which name you want us to use.

Thank you.  No need to add me to the contributor's list.

PR #207, opened 2024-01-15
Buffer for more Speed

tg-freigmbh:
I realized that running tests using ant was much slower than running them from the IDE. The culprit was the legacy-xml reporter.
grafik
Using an unbuffered Output is very slow, even if you are using an .m2 disk..

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/LauncherSupport.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/LauncherSupport.java
@@ -45,7 +45,15 @@ import org.junit.platform.launcher.core.LauncherFactory;
 import org.junit.platform.launcher.listeners.SummaryGeneratingListener;
 import org.junit.platform.launcher.listeners.TestExecutionSummary;
 
-import java.io.*;
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.io.PrintStream;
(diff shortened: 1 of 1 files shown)


jaikiran:
Hello @tg-freigmbh, very good catch. This indeed is an issue. I've added a trivial comment inline, other than that this looks good.

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/LauncherSupport.java
```diff
@@ -45,14 +45,7 @@
...
 
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.PipedInputStream;
-import java.io.PipedOutputStream;
-import java.io.PrintStream;
-import java.io.UncheckedIOException;
+import java.io.*;
```
jaikiran, line 48: Can you undo the `*` import and use the previous existing imports plus one new for the BufferedOutputStream?

tg-freigmbh:
I tried, but my ide realy doesnt like that :D maybe I can do it manually on github

src/main/org/apache/tools/ant/taskdefs/optional/junitlauncher/LauncherSupport.java
tg-freigmbh: done

jaikiran:
Thank you for the update, looks good to me. Since you are a first time contributor to Ant, we add your name to the contributors list we maintain in the repo. Is there a specific first name, last name you want us to add you as?

tg-freigmbh:
Ah I already made an Contribution  under a different account, I am Gösen ;)

PR #208, opened 2024-02-21
fix <compilerarg> example documentation in manual

mebigfatguy:
documentation says the attribute name for `<compilerarg>` is arg, but it's value. Updated doc.


jaikiran:
Thank you Dave for this fix. The change looks correct to me and matches the `compilerarg` documentation of the `javac` task.

PR #213, opened 2024-10-03
BUG-ID : 59160 , fix extracting host for ipv6 address

FROZENCADET:
Consider the last occurrence of ':' while getting the indexOfPath as IPv6 address has more than one ':'.

### DIFF
--- a/src/main/org/apache/tools/ant/taskdefs/optional/ssh/Scp.java
+++ b/src/main/org/apache/tools/ant/taskdefs/optional/ssh/Scp.java
@@ -437,7 +437,7 @@ public class Scp extends SSHBase {
 
         // For IPv6 address having more than one ":", it should look for the last occurrence
         final int indexOfPath = uri.lastIndexof(':');
-        if (indexOfPath == -1) {
+        if (indexOfPath <= indexOfAt) {
             throw new BuildException("no remote path in %s", uri);
         }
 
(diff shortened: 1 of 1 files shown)


src/main/org/apache/tools/ant/taskdefs/optional/ssh/Scp.java
```diff
@@ -435,7 +435,8 @@ private String parseUri(final String uri) {
             throw new BuildException("no username was given.  Can't authenticate.");
         }
 
-        final int indexOfPath = uri.indexOf(':', indexOfAt + 1);
+        // For IPv6 address having more than one ":", it should look for the last occurrence
+        final int indexOfPath = uri.lastIndexof(':');
         if (indexOfPath == -1) {
```
bodewig, line 440: we should still ensure `indexOfPath` is bigger than `indexOfAt`. So maybe we change the if to not only throw if it is -1 but rather if it is less than or equal `indexOfAt` (equal could be the case if `indexOfAt` is -1 itself.

FROZENCADET: I agree.. have made the required changes

bodewig:
@FROZENCADET actually the code didn't even compile see 2540910de

Merged, thank you.

PR #215, opened 2024-12-08
Upgrade Hamcrest to 3.0

tomodachi94:
If upstream changelogs are to be believed, there are no breaking changes except for:
* The packaging method being changed[1] from several JARs to one JAR.
* An annotation, `org.hamcrest.Factory`, being removed (we don't use it, and upstream says that people should not have been using it externally)
* The minimum version being upgraded to Java 8[2], which is already the case in Ant.

[1]: https://github.com/hamcrest/JavaHamcrest/releases/tag/v2.1
[2]: https://github.com/hamcrest/JavaHamcrest/releases/tag/v3.0


bodewig:
Things seem to work with Hamcrest 3, thank you.

Actually the hamcrest jars are checked in into the repository (something that look strange for people today, I know :-), so you need to remove the 1.x jars and add the 3.0 one. And https://github.com/apache/ant/blob/master/lib/optional/LICENSE.hamcrest.txt should be adjusted as well.

tomodachi94:
Pushed both changes as new commits onto this PR's branch. Let me know if you prefer everything to be squashed into one commit.

bodewig:
thank you

Do you want to be credited inside our CONTRIBUTORS file - and if yes, by which name?

bodewig:
> I'm going to take a look at the issue tracker to see if there's anything there that I can help with.

Great. We are not always the quickest people to respond, please don't allow this to discourage you.

PR #217, opened 2025-03-13
Fix PermissionUtils.FileType.of for sym links

kiendang:
Currently `PermissionUtils.FileType.of` returns `REGULAR_FILE` instead of `SYMLINK` for sym links.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -244,6 +244,7 @@ Kevin Greiner
 Kevin Jackson
 Kevin Ross
 Kevin Z. Grey
+Kien Dang
 Kim Hansen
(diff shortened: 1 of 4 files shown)


kiendang:
Changes made 👍 
> but for Ant there will probably never be a real difference
Yup I ran into this while using the zip tools for something outside of Ant.

bodewig (approved):

bodewig:
Thank you

bodewig:
In general I would not recommend using Ant's zip classes but rather Apache Commons Compress unless you've got a dependency on Ant anyway. Commons Compress has been seeded with code written for Ant and expanded from there.

PR #218, opened 2025-05-12
Improve JUnit report title handling

Vampire:


bodewig:
Thank you @Vampire

By which name do you want to be credited in the CONTRIBUTORS file?

bodewig:
> My actual name will do.

fine. This is not always the case, so we tend to ask :-)

Thanks again.

Vampire:
Sure, no problem, thanks for the attribution. :-)

PR #220, opened 2025-06-03
Update the SPDX ID of license

Goooler:
Seems the current used `The Apache Software License, Version 2.0` is outdated, and it can't be recognized by Maven Central.

See https://central.sonatype.com/artifact/org.apache.ant/ant/1.10.15

image

### DIFF
--- a/src/etc/poms/pom.xml
+++ b/src/etc/poms/pom.xml
@@ -31,7 +31,7 @@
   <description>master POM</description>
   <licenses>
     <license>
-      <name>The Apache Software License, Version 2.0</name>
+      <name>Apache-2.0</name>
       <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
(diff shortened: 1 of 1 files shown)


paulk-asert:
I am not sure it is the text in that field that is the problem. I actually just don't see the license or license tags appear at all in the ant pom. Perhaps those need to be added to the file here:

https://github.com/apache/ant/blob/master/src/etc/poms/ant/pom.xml

The parent pom seems to show the current wording still:
Screenshot 2025-06-03 at 9 00 24 pm

Goooler:
Seems license nodes are missing for all POM under https://github.com/apache/ant/tree/master/src/etc/poms, I'm not sure if we should append for them or not...

bodewig:
I would expect the `licenses` to be inherited from https://github.com/apache/ant/blob/master/src/etc/poms/pom.xml

PR #221, opened 2025-06-08
differenciate between errors and failures for junitlauncher

bodewig:
Try to tell errors from failures in junitlaucher's formatters.

@Vampire and @jaikiran I'd like to have a second pair of eyes since I'm not familiar with the JUnit5 API at all.

### DIFF
--- a/WHATSNEW
+++ b/WHATSNEW
@@ -32,6 +32,10 @@ Fixed bugs:
    path even if it was not a file separator.
    Bugzilla Report 69680
 
+ * <junitlauncher>'s legacy formatters now separate errors from
+   failures like their <junit> counterparts did.
+   Bugzilla Report 69687
+
 Other changes:
 --------------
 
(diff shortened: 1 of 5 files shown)


Vampire:
Hm, yeah, JUnit Platform indeed does not differentiate between failure and error anymore.
I have no idea why, maybe @marcphilipp could shed some light on why this is not done anymore.

Gradle indeed does also not do this differentiation yet, but reports both cases as failure.

IntelliJ IDEA, when running the tests itself and not through Gradle delegation, does do the separation.

Seems like a big cross-project inconsistency mess. :-(

For me as user it was majorly confusing, that errors are reported as failures when using Ant, while when using IntelliJ IDEA integration they were shown as error.

So maybe it is appropriate to follow suite with JUnit logic and not differentiate them.
But then at least your own consumers of the file (the XSLs) should expect that it might be missing, as the generated HTML report looks broken if there is not even an errors attribute.

marcphilipp:
> Hm, yeah, JUnit Platform indeed does not differentiate between failure and error anymore. I have no idea why, maybe @marcphilipp could shed some light on why this is not done anymore.
> 
> Gradle indeed does also not do this differentiation yet, but reports both cases as failure.

Since both are signaled via exceptions, we decided to simplify things when we started work on JUnit 5 (or JUnit Lambda as it was called back in the day). If you want to differentiate "failures" from "errors", you can check whether the `Throwable` is an instance of `AssertionError`.

PR #222, opened 2025-06-27
Exit ant.bat with ERROR_FILE_NOT_FOUND if ANT_HOME isn't/can't be set

robho:
If ant.bat can't run because ANT_HOME isn't set correctly I think the script should exit with an error code. ERROR_FILE_NOT_FOUND (2) seems reasonable to me.


robho:
I have changed to use ERROR_ENVVAR_NOT_FOUND.

There wasn't much thought behind choosing ERROR_FILE_NOT_FOUND and I agree that ERROR_ENVVAR_NOT_FOUND does seem more appropriate.

bodewig:
Many thanks @robho - do you want us to credit you in CONTRIBUTORS and contributors.xml? If so which name do you want us to use?

PR #223, opened 2025-07-26
Use StreamSupport and spliterator to create a stream for ResourceCollection

mattbertolini:
I came across this while working on a custom Ant task. I saw what I think is the ResourceCollection stream() method eagerly loading elements when creating a stream. Not sure if that was intentional or not. If it is, feel free to close the PR.  If not, I've created this to update it to use the StreamSupport factory method and a spliterator. I think this is how the JDK Collection creates the stream. This will make using things like findFirst a bit faster as it won't have to load every element to find the first.

### DIFF
--- a/src/main/org/apache/tools/ant/types/ResourceCollection.java
+++ b/src/main/org/apache/tools/ant/types/ResourceCollection.java
@@ -14,6 +14,7 @@
 package org.apache.tools.ant.types;
 
 import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 /**
  * Interface describing a collection of Resources.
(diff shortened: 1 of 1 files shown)


bodewig:
Thank you.

At first glance I don't believe there's been a reason to not use StreamSupport when the [big commit](https://github.com/apache/ant/commit/b7d1e9bde44cb8e5233d6e70bb96e14cbb2f3e2d) that moved us to Java 8 features was created.

But then I was wondering whether we should override `spliterator` since we know `size()` and could do better than use the default. Which reminded me that some implementations don't know their size without iteration (filtered collections of any kind), Those collections use an internal cache to avoid repeated iterations (`MappedResourceCollection` for example). The default `spliterator` uses `iterator` if I understand correctly, so using StreamSupport the way you do should also be using the cache.

bodewig:
> I thought that "big commit" sounded familiar! 😀 Matt

sorry :-)

bodewig:
Thank you @mattbertolini , I'll try to find some time during the weekend to look through our implementations. Maybe, just maybe, it would be better to define `iterator` only in those subclasses that have a better idea of their size. I really need to look through the code to get a better feeling for it.

mattbertolini:
@bodewig Thank you for the detailed analysis. It was definitely more complex than I was originally thinking. No worries on the extra work. I'm happy to do it. I've reverted the spliterator change and have the original stream change back.

PR #233, opened 2026-05-24
Cyclonedx

bodewig:
Example SBOMs can be seen at https://repository.apache.org/content/repositories/orgapacheant-1069/org/apache/ant/ - of course I don't intend to publish anything at this point.


sboms/ant-1.10.18alpha.cdx.json
```diff
@@ -0,0 +1,355 @@
...
+              "type" : "vcs",
+              "url" : "https://gitbox.apache.org/repos/asf/ant-antlibs-cyclonedx.git"
+            },
+            {
+              "type" : "license",
+              "url" : "https://www.apache.org/licenses/LICENSE-2.0.txt"
+            },
+            {
+              "type" : "build-system",
+              "url" : "https://ci-builds.apache.org/job/Ant/job/CycloneDX%20Antlib/"
```
jaikiran: Hello Stefan, is the "build-system" type mandated for these SBOM files? Although `ci-builds.apache.org` is something that is almost guaranteed to be present (at least in near future). I don't know if we can guarantee that the jobs referred here are going to be "permanent". We keep changing their names etc... as and when it pleases us because they are just internal jobs. So I'm unsure if we should publish this job link in the SBOM.

sboms/ant-1.10.18alpha.cdx.json
```diff
@@ -0,0 +1,355 @@
...
+          "type" : "release-notes",
+          "url" : "https://github.com/apache/ant/blob/master/WHATSNEW"
+        },
+        {
+          "type" : "security-contact",
+          "url" : "https://www.apache.org/security/"
+        },
+        {
+          "type" : "rfc-9116",
+          "url" : "https://ant.apache.org/.well-known/security.txt"
```
jaikiran: I didn't know that there was a RFC which specifies a security.txt file for projects. Good to know.

bodewig: actually the "type" is an enum defined by the spec: https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_externalReferences_items_type - the types may be more restricted based on the context they appear in as they may apply to the SBOM itself or a component mentioned therein.

I vaguely recalled the RFC as I do have security.txt files for my own websites as well, but had forgotten it was an actual RFC.

sboms/ant-1.10.18alpha.cdx.json
bodewig: no external link is mandatory at all. I thought I'd put in as many as we can reasonably fill.

What you say about the CI system may also be true for the Bug-Tracker. I could change things to stop at the level of the Ant project rather than the individual component. Or even at the system level itself (i.e. ci-builds.apache.org and issues.apache.org).

The concrete line you commented on comes from the released ant-cyclonedx library and can only be modified with a new release, but I believe we are talking about the links in general.

jaikiran: Hello Stefan, for issue tracker, mailing list links, I think having a very precise link with the project id in it is worth it. So what you have for those right now looks good to me. To me those links are more user facing because they are used by actual users to either initiate a discussion or file issues. Plus the backing infrastructure for those (like the Ant project in the issue tracker) is much more official.

On the other hand, the CI jobs link pointing to the exact job, I feel is a bit adhoc. And I don't think many users actual would be using that link. So for the `build-system` type, I think it might be OK to either leave that type out completely or just point to the top level `https://ci-builds.apache.org`.

jaikiran:
Apart from the trivial part about the build-system links in the SBOM file, the rest of this looks good to me.

sboms/ant-1.10.18alpha.cdx.json
bodewig: -> e72bf3ade

this is fine with me. I'll also remove the links from the Antlibs and Ivy while I'm at it.

bodewig:
Thank you @jaikiran . I'll remove the sample SBOMs and then merge this.

